PR #11, opened 2020-05-28
Add heroku deployment dependencies

### LINKED ISSUE #2: Set up the shared deployment environment
We'll need to decide how and where to deploy the app while we are developing it. You will probably set up a local environment on your own machine for the purposes of developing and debugging. What we also need is somewhere me & Tim could look at the app, test it and provide feedback. It is essential to be able to deploy multiple versions at the same time (e. g. one version per branch), and it would make sense to make the deployment completely automated.

We're currently looking at setting this up on the EBI infrastructure. Could you please provide a list of requirements which you need in terms of the web server, database, and your preferred method of setting up the continuous integration?

joj0s:
This PR includes all the dependencies for deploying to Heroku as well as instructions on how to do so. Closes #2 .The instructions assume that a Heroku account has been set up.

1. Create a new Heroku app via the dashboard
2. Once created, navigate to the "Deploy" tab and in the "Deployment Method" menu, select "GitHub".
3. Search for the "trait-curation" repo and click "Connect"
4. Below, in the Automatic deploys menu, select the master branch. (I would suggest leaving the "Wait for CI" box unchecked until we set up the CI)

The app should now be successfully deployed. Follow-up instructions are included in order to enable Review Apps, which automatically create a deployment instance for every new PR.

1. Navigate to the Deploy tab, in the "Add this app to a pipeline" menu.
2. Choose "Create new pipeline" and name it the same as the project
3. Moving to staging or production shouldn't matter given that this is a development environment 
4. We now have to connect the whole pipeline to GitHub as well. Navigate to Settings > Connect to GitHub and select the repository
5. Once the pipeline is created and connected to the GitHub repo, the left most column should have a "Enable Review Apps" button
5. Click it, select "Create new review apps for new pull requests automatically " and make sure to select the "Create new review apps for new pull requests automatically" option
6. Optionally, friendlier URL patterns for review apps can be chosen by navigating to Settings > Review Apps


joj0s:
Update: I am not sure whether we should run the database migrations via the Procfile command (this is how it is done in this commit) or via the bash console provided by Heroku. I would say we follow the current approach and if any migrations fail we can run them manually. In the testing app the migrations ran fine

tskir:
The deployment instructions you provided in the PR description are excellent. To make them permanent (since the PR description isn't—it will get buried when we close it), please add them as a separate Markdown file in the `docs/` subdirectory.

As an example of what I'm talking about, this is how we have docs organised for the main Open Targets pipeline: https://github.com/EBIvariation/eva-cttv-pipeline/tree/master/docs

tskir:
Regarding the migrations, I don't have much experience with it. From what you describe, it sounds like the Procfile approach is automated, and we'll have to run migrations manually only if they fail; and the alternative bash console approach is fully manual. If I understood that correctly, then let's go with the (mostly) automated approach.

traitcuration/settings.py
```diff
@@ -118,3 +118,6 @@
 # https://docs.djangoproject.com/en/3.0/howto/static-files/
 
 STATIC_URL = '/static/'
+
+# Activate Django-Heroku.
+django_heroku.settings(locals())
```
tskir, line 123: PEP8 suggests that all files should add with a newline, could you please add that? Also for `Procfile`

joj0s:
I have added newlines at the files requested. I also enabled automatic linting in my environment to prevent this from happening in the future. Now, i excluded my vscode settings from git, but if you also use it and you want to keep the settings uniform, let me know and I can add them back.

docs/heroku-deployment.md
```diff
@@ -0,0 +1,19 @@
+
+Step by step guide on how to deploy the app in Heroku and enable automatic pull request deployment as well.
```
tskir, line 2:
```suggestion
# Setting up deployment in Heroku and automated pull request deployment
```

docs/heroku-deployment.md
```diff
@@ -0,0 +1,19 @@
...
+The instructions assume that a Heroku account has been set up.
+
+
```
tskir, line 5:
```suggestion

```
Extra blank line

traitcuration/settings.py
```diff
@@ -9,7 +9,7 @@
 For the full list of settings and their values, see
 https://docs.djangoproject.com/en/3.0/ref/settings/
 """
-
+import django_heroku
```
tskir, line 12:
```suggestion

import django_heroku
```

Adding a blank line between the end of docstring and the first statement

docs/heroku-deployment.md
tskir: Removing blank line from the start of the file and converting the first sentence into a header

PR #13, opened 2020-06-04
Create data documentation md file

### LINKED ISSUE #7: Finalise the complete list of trait statuses and transitions between them
The Google doc included a simplified example of this. Now we need to discuss & finalise this. We need to make sure that the trait status list/transitions fully take into account:
* Terms becoming obsolete
* Terms becoming deleted from the ontology
* Terms having none/partial/full reviews

### LINKED ISSUE #8: Describe and discuss the data model
It was mentioned during the last call that we should have a shared Google doc for discussing and finalising the data model. Could you please create this doc and link it from here?

joj0s:
This PR contains a markdown file containing documentation for the data model as well as the trait status transitions, along with a mermaid diagram and its source code. Closes #7 , closes #8 

A consideration I have is that the table describing the trait statuses is very big and therefore the markdown required to output it is quite not user-friendly, so we might want to change the way we show that piece of information.

I also uploaded the png diagram to Imgur so that is stays permanently and that the link is small


docs/data-model.md
```diff
@@ -0,0 +1,78 @@
+# Data model description and trait status explanation
+
+- `traits` (id, trait_name, current_mapping[foreign_key], status[computed_key], number_of_source_records, timestamp_added, timestamp_updated)
+
+The master status is meant to reflect the state of the mapping and provide insight as to what should be the next action. It can contain all the possible trait statuses, as well as `unmapped` and in `in_review` statuses. It is never changed explicitly and is always computed based on:
```
tskir, line 5: Starting a separate paragraph in this context confuses Markdown, and causes it to break the list structure. I think it's better to restructure this so that column definitions are first-level items, and all comments and explanations are second (and so on) levels, something like this:

```markdown
- `traits` (id, trait_name, ...)
  + The master status is meant to...
- `users` (...)
- `mappings` (...)
  + The `is_reviewed` field status is...
```

docs/data-model.md
```diff
@@ -0,0 +1,78 @@
...
+| A batch PR is submitted for this and other terms (for import and creation)                                                                                                   | awaiting_creation    | reviewed                                                        | awaiting_creation                                                |
+| After the PR is closed and new terms are created, the curator <br>manually selects a new term                                                                                | current              | not_reviewed                                                    | awaiting_review                                                  |
+| A necessary number of reviews for the newly created term is accumulated                                                                                                      | current              | reviewed                                                        | current                                                          |
+
+### Diagram:
+
+![](https://i.imgur.com/tUBBipu.png)
+
+Mermaid diagram source code :
```
tskir, line 52: It's best to move this into a separate file, like `docs/statuses_diagram.mmd` (`.mmd` is [the official extension](https://github.com/mermaid-js/mermaid/issues/147) for Mermaid source code).

docs/data-model.md
```diff
@@ -0,0 +1,78 @@
...
+| A batch PR is submitted for this and other terms (for import and creation)                                                                                                   | awaiting_creation    | reviewed                                                        | awaiting_creation                                                |
+| After the PR is closed and new terms are created, the curator <br>manually selects a new term                                                                                | current              | not_reviewed                                                    | awaiting_review                                                  |
+| A necessary number of reviews for the newly created term is accumulated                                                                                                      | current              | reviewed                                                        | current                                                          |
+
+### Diagram:
+
+![](https://i.imgur.com/tUBBipu.png)
```
tskir, line 50: It's better to commit the PNG directly into repository as a binary file than to rely on an external resource (even though, admittedly, Imgur is much better than hotlinking Mermaid servers directly).

While normally committing binary files (especially large) is bad, for documentation PNGs it's perfectly fine.

docs/data-model.md
```diff
@@ -0,0 +1,78 @@
...
+| ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------- | --------------------------------------------------------------- | ---------------------------------------------------------------- |
+| **Case 1. Curator chooses an existing term from EFO**                                                                                                                        | current              | not_reviewed                                                    | awaiting review                                                  |
+| A necessary number of reviews is accumulated                                                                                                                                 | current              | reviewed                                                        | current                                                          |
```
tskir, line 35: Regarding the Markdown layout for this table (since you've raised concerns about it in the PR description): it's mostly fine as it is; this table is unlikely to be frequently changed, so the most important thing is that it should be readable in the final rendered Markdown, which it is.

However, if you want to improve readability of the source code as well, I can suggest not trying to keep column delimiters aligned. Markdown allows for this, and in this case it will look prettier. Compare, before:

```markdown
| A necessary number of reviews is accumulated                                                                                                                                 | current              | reviewed                                                        | current                                                          |
```

And after:

```markdown
| A necessary number of reviews is accumulated | current | reviewed | current |
```

tskir:
Great, all comments addressed without further issues, merging this

Issue #14, opened 2020-06-04
Implement import of trait names from ClinVar


tskir:
Information about traits can be retrieved from a static, periodically updated endpoint https://ftp.ncbi.nlm.nih.gov/pub/clinvar/tab_delimited/variant_summary.txt.gz. There is an accompanying checksum file https://ftp.ncbi.nlm.nih.gov/pub/clinvar/tab_delimited/variant_summary.txt.gz.md5.

The main file is a gzipped TSV containing about two dozen columns (first line, starting with `#`, is a header). There are three columns of interest:
* **PhenotypeList** column contains trait names. There can be one or multiple per record. If there are multiple, they are separated with semicolons, for example _Inborn genetic diseases;Mitochondrial complex 1 deficiency, nuclear type 21;Mitochondrial complex I deficiency;not provided_
* **RCVaccession** contains IDs of ClinVar RCV records. An RCV record, in general, associates multiple genetic variants with multiple trait names. Similarly, there can be one or multiple, for example _RCV000622708;RCV000735415;RCV000000017;RCV000196589_.
* **AlleleID** contains the identifier of a genetic variant. Example: _15046_.

These columns follow this list of constraints:
* The number of semicolon-separated values in **PhenotypeList** and **RCVaccession** is always the same for a given record and is always 1 or greater. The values with same same index form pairs.
* **AlleleID** always contains an integer, however, it is _not_ a key for the table and several rows can have the same AlleleID (see below).

Trait names imported into the database should represent a unique set of all trait names present in this file.

To calculate the number of records assigned to each trait, we need to consider all tuples of (AlleleID, RCVaccession, PhenotypeList). For example, exploding along RCVaccession and PhenotypeList, the sample record described above would generate four tuples:
* (15046, RCV000622708, Inborn genetic diseases)
* (15046, RCV000735415, Mitochondrial complex 1 deficiency, nuclear type 21)
* (15046, RCV000000017, Mitochondrial complex I deficiency)
* (15046, RCV000196589, not provided)

When we do this for all rows in the file, we need to combine all tuples together and deduplicate them (leave only unique tuples). In this final set, the number of tuples which mention a given trait name is the number of records associated with this trait, which should also be imported into the database and stored in a separate field.

[...]

PR #35, opened 2020-06-08
Prototype

### LINKED ISSUE #3: Implement minimal prototype for the app
This can be flexible depending on how you think it could be implemented best. Some ideas:
* The prototype should include minimal UI and data model
* It may have only one page, for instance the list of traits and to which term they are mapped to
* It doesn't need to include any functionality on loading traits from an external source, just using dummy data will be OK for now

### LINKED ISSUE #4: Deploy & test the minimal prototype
Depends on #2 and #3.

joj0s:
This PR includes the first prototype of the app, featuring the first implementation of the data model, as well as a table showing all the inserted data, and a page for each trait showing its details. Closes #3, closes #4 .

I'd like to draw the attention to the data model file (`traitcuration/traits/models.py`). In particular these points:

- The *on_delete* behaviour of the different models and whether they can ever be deleted or not. Right now I am treating them as such:
  + In the trait model, if its current mapping is deleted I am setting it to null. If we want to never delete a mapping from the database, the other option would be to protect it from being deleted
  + I am protecting traits as well as ontology terms  being deleted
  + With the current implementation, users can be deleted, and when they are, their reviews and comments are set to a null owner
- Also regarding the user model. I have created a new one, however Django already has a built in user model for authentication. I suggest we use the custom one for testing, and when we implement actual authentication we could see if it is more convenient. Let me know if you agree
- I would also appreciate if you could check the individual fields in the model file and see if any field that should be nullable isn't, and vice versa. (In Django, `null` means it can be empty in the database, and `blank` means that that field can be left empty when filling in Django forms. )


joj0s:
The PR has been automatically deployed here https://clinvar-trai-prototype-pmia54t.herokuapp.com/ . You can create and store new data in the database as you please by going to `/admin` and logging in with **admin** username and **admin** password.

traitcuration/static/styles/layout/header/_header.scss
```diff
@@ -0,0 +1,25 @@
+@import "../../utilities/variables";
+
+.header {
+  background-color: $color-primary !important;
+  padding: 0 15px;
```
tskir, line 5: I think there are two competing (but also complementary) approaches to use either `px` (absolute) or `rem` (relative) values for sizes of elements. It looks like there are benefits and drawbacks for all of them (I found a good summary here: https://stackoverflow.com/questions/11799236/should-i-use-px-or-rem-value-units-in-my-css), so I'll leave it up to you which way to choose. My understanding that currently we're going with pixels

traitcuration/traits/models.py
```diff
@@ -0,0 +1,66 @@
+from django.db import models
+
+
+class Trait(models.Model):
+    name = models.CharField(max_length=200)
```
tskir, line 5: If ClinVar one day introduces a trait name longer than 200 characters, what will happen under the current model? I'm guessing we should introduce a check into the import process (once that is implemented) to avoid a hard crash

traitcuration/traits/models.py
```diff
@@ -0,0 +1,66 @@
...
+    current_mapping = models.ForeignKey(
+        'Mapping', on_delete=models.SET_NULL, null=True, blank=True)
```
tskir, line 7:
```suggestion
    current_mapping = models.ForeignKey('Mapping', on_delete=models.SET_NULL, null=True, blank=True)
```

This fits on one line in less than 120 characters

traitcuration/traits/models.py
```diff
@@ -0,0 +1,66 @@
...
+    status = models.CharField(max_length=50)
```
tskir, line 8: Just to clarify: the status is a `CharField` only because this is a prototype, right? In the future it would have to become a `ComputedField`

tskir: Also I just want to make sure that either there is already a ticket for implementing this part in the backend (in this case it should be linked under this comment), or not yet (in which case I'll create one)

traitcuration/traits/models.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __str__(self):
+        return self.username
+
+
+class OntologyTerm(models.Model):
+    curie = models.CharField(max_length=50, blank=True, null=True, unique=True)
+    uri = models.URLField(null=True, blank=True, unique=True)
+    label = models.CharField(max_length=200)
+    status = models.CharField(max_length=50)
```
tskir, line 30: This should also (eventually) be not a char, but some sort of enum

traitcuration/static/js/uikit/uikit-icons.js
```diff
@@ -0,0 +1,167 @@
+/*! UIkit 3.5.3 | https://www.getuikit.com | (c) 2014 - 2020 YOOtheme | MIT License */
```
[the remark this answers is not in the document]
joj0s: I have mixed feelings about CDNs to be honest, mainly about security issues, although I could look at setting up referrer policies. On the other hand, I'm pretty sure using Yarn with Django will still require vendoring them, although I can search for a pip package to possibly solve that.

traitcuration/static/styles/layout/header/_header.scss
joj0s: In my experience, em for font-sizes and px for the rest works well and responsively. Of course if any problem about that arises I can tweak the CSS as needed

traitcuration/static/js/uikit/uikit-icons.js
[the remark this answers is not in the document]
joj0s: https://github.com/epineda/django-yarn

I found that, I will try it out and see how well it works

traitcuration/traits/models.py
joj0s: Yes, I can add that check. Should I increase the maximum length as well?

traitcuration/traits/models.py
joj0s: Yes that is so I can manually add a status via the Django Admin interface, but I have been thinking about that. There is no built-in `ComputedField` option in the Django ORM. There are two solutions to that. 
- Keep it as a CharField and add a check whenever a review is issued or whenever a mapping is changed to automatically compute the field's value
- Use this package https://pypi.org/project/django-computedfields/

traitcuration/traits/models.py
joj0s: It can be converted into an enum by adding a `choices` attribute to the field. I will add that in the implemented changes

traitcuration/static/js/uikit/uikit-icons.js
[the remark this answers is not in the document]
joj0s: Actually it turns out Django does take care of security issues as far as I understand. So I would say since we are not going to use a lot of static dependencies, we can use CDNs to remove the overhead of static files

traitcuration/settings.py
```diff
@@ -117,8 +124,26 @@
 
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/3.0/howto/static-files/
+COMPRESS_ENABLED = True
+
+
```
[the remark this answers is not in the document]
joj0s: You are right, I will remove that.

joj0s:
I pushed a commit addressing all your comments besides the ones open for discussion. In particular:
- Added loading for UIKit via CDN
- Turned the status fields into enum (discussion on their computing is pending)
- Removed double blank line from settings.py
- Made the code in models.py, line 6, single line

I also added the model for reviews, which I accidentally left off of the previous commits

traitcuration/static/styles/layout/header/_header.scss
joj0s: Actually I have given it some more thought. I decided to use rem globally (except for box-shadow properties) to maintain consistency across the CSS code. I think however that we should first merge the open PRs and then I can change it so that I can avoid as many merge conflicts as possible. Do you agree?

traitcuration/static/js/uikit/uikit-icons.js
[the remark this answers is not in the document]
tskir: > I have mixed feelings about CDNs to be honest, mainly about security issues

Yeah, that's true, I realise CDNs have their pros and cons. If we were developing something handling highly sensitive data, then maybe I wouldn't go for them. But public CDNs are supposed to be reasonably safe, and I feel this is the case where simplicity and speed of development are more important.

So, TL;DR, I support use of CDNs.

traitcuration/static/styles/layout/header/_header.scss
tskir: Yes, absolutely, this sounds sensible. I've created a separate issue to handle this: https://github.com/EBIvariation/trait-curation/issues/40

If you eventually decide that it's not worth it, then this is also fine. In this case just close that issue without changing anything.

traitcuration/traits/models.py
tskir: No, I think the current lenght limit is fine. If ClinVar actually adds a trait which name is monstrously large, then I _want_ the import to fail and to notify us in this way, because who knows what they can put in place of a trait name :)

Since this PR only concerns the prototype, I've added a separate checkbox to #14 description to implement this check

traitcuration/traits/models.py
tskir: @joj0s As far as I see, this hasn't been resolved yet. Granted, this is tiny, but if we're following 120 character line limit, then we might as well be consistent :)

traitcuration/traits/models.py
tskir: I see. This looks like it can be a whole separate issue, I've created one: https://github.com/EBIvariation/trait-curation/issues/41. Let's use it to evaluate all those possible approaches.

traitcuration/traits/models.py
```diff
@@ -0,0 +1,91 @@
...
+    label = models.CharField(max_length=200)
+    status = models.CharField(max_length=50, choices=[
+        ('current', 'Current'),
+        ('obsolete', 'Obsolete'),
+        ('deleted', 'Deleted'),
+        ('unmapped', 'Unmapped'),
+        ('needs_import', 'Needs Import'),
+        ('awaiting_import', 'Awaiting Import'),
+        ('needs_creation', 'Needs Creation'),
+        ('awaiting_creation', 'Awaiting Creation')])
```
tskir, line 47: To avoid code duplication, I think this list should be moved to a variable, and then the list of statuses for the Trait should be this variable _plus_ "awaiting review".

traitcuration/static/styles/layout/header/_header.scss
tskir: Yes, absolutely, this sounds sensible. I've created a separate meta-issue to track cases like this, when there are tiny things to change and it doesn't make sense to postpone merging otherwise good PRs: https://github.com/EBIvariation/trait-curation/issues/40

PR #36, opened 2020-06-14
Filters for text searching and statuses

### LINKED ISSUE #16: Implement text filter
It should perform free-text search in "Trait name", "Term" and "Label" columns.

image

### LINKED ISSUE #17: Implement filtering by status
This should contain radio buttons, allowing to select only one value at a time.

image

This is related to #16. For both of those issues, the acceptable current implementation is to do nothing when either the text or status selection changes, _until_ the user have clicked the Filter button. We can implement instant filtering later on. However, if you find that it's easier to implement instant filtering from the start then it's also fine.

### LINKED ISSUE #20: Implement trait view table
This is the main table which displays the traits.

image

joj0s:
This PR adds a text filter on top of the "Browse Traits" page that filters by trait name, ontology labels and ontology CURIE. It also adds buttons for selecting each distinct status of a trait. 

Compared to the UI mockup, I excluded the "FIlter" button next to the textbox since it is all done instantly using JavaScript and we don't need a button to trigger a request.

Closes #16, closes #17, closes #20


joj0s:
@tskir  I have also added colored labels to the trait table and links to to point the CURIE column to its appropriate URI. if you like the changes, we can also close #20

traitcuration/static/js/table-filtering.js
```diff
@@ -0,0 +1,33 @@
...
+  valueNames: ["name", "curie", "label", "status"],
+};
+
+var table = new List("traits-table", options);
+
+function search(searchstring) {
+  table.search(searchstring, ["name", "curie", "label"]);
+}
+
+function filterClicked(status, activeClass) {
```
tskir, line 11: Could you add some documentation to this script? I'm thinking one top-level description (like a docstring) + some comments throughout the code, at least for the major branching points. They don't have to be verbose

traitcuration/traits/views.py
```diff
@@ -0,0 +1,33 @@
...
+    return render(request, 'traits/browse.html', context)
+
+
+def trait_detail(request, pk):
+    trait = get_object_or_404(Trait, pk=pk)
+    context = {"trait": trait}
+    return render(request, 'traits/trait_detail.html', context)
+
+
+def get_status_dict(traits):
```
tskir, line 18: So concerning the `views.py` module, do you suppose it would be better to leave only the actual views in it (e.g. something which returns `render(...)`, and move helper routines, such as `get_status_dict`, into a separate module?

traitcuration/static/js/listjs/list.js
```diff
@@ -0,0 +1,2 @@
+// List.js v1.4.1 (http://www.listjs.com) by Jonny Strömberg (http://javve.com)
```
tskir, line 1: The same comment applies to this library, depending on which decision we will go with in #35: either use external CDN, or install via Yarn, or keep locally and include update/installation instructions

joj0s:
> The only small bug is that the behaviour of the Enter button is somewhat confusing: when a user presses it while focused on the search field, anything the user has entered there will get erased and they will be redirected to the front page. Enter should not do this.

You are right, I didn't test the Enter button functionality. I will fix that
 
>Another consequence of instant search is that it's not possible to bookmark a combination of specific filters, as there isn't a dedicated URL to use for that. If this can be fixed easily, then please do so as part of this PR. Otherwise, we can create a separate low-priority issue in the backlog for that.

Yes, that is the drawback of using JavaScript to filter. We cannot include the filters in a URL so I think we should add a new issue for that, although I am not quite sure what that would be. Possibly a page with query history?

traitcuration/traits/views.py
joj0s: Yeah, I also thought about it and it would be cleaner. I moved it into a `utils.py` module

joj0s:
Regarding the second iteration of changes in this PR:

- Created a `utils.py` file and moved the helper function from views.py to there
- Documented the JS script for filtering
- Fixed the Enter button functionality
- Added loading list.js library via CDN

Issue #32, opened 2020-06-06
Design a page to display status & trigger external imports


tskir:
There are three external datasources which the app will communicate with:
* Importing trait names from ClinVar #14 
* Obtaining possible mappings from ZOOMA #30 
* Querying OLS to update ontology term status #31 

This page will be essentially an extension of the "External datasources" page which you added in #39. Here are my thoughts on what we could include on that page:
* Buttons for triggering ClinVar, OLS, ZOOMA imports
* Configuration variables:
  + URL for ClinVar download (currently https://ftp.ncbi.nlm.nih.gov/pub/clinvar/tab_delimited/variant_summary.txt.gz)
  + How frequently to trigger re-imports of 
  + Base URLs for OLS and ZOOMA
* Status of import for all three datasources
  + If a particular import is currently in progress, display "In progress"
  + If failed, display "Failed"
  + Otherwise, display the timestamp on when the import was last performed
* Development options:
  + Button for purging the entire database
  + A number of rows to import from `variant_summary.txt.gz` (to avoid importing the full set, which may take a lot of time and run over Heroku free limits)

tskir:
Since this page seems handy to have, including for development/debugging purposes, I've raised the priority to high. I have also amended the description of this issue to include some things which can be added to this page.

joj0s:
https://www.figma.com/file/pd5DHJSsldVtJcbp4RvkPb/GSoC-Trait-Mapping-App?node-id=0%3A1
Here is a prototype of this page that I created. I haven't included development button such as "Import dummy data" or "Purge database" since this is supposed to be the production prototype, so I can add those at the top or bottom of the page.

tcezard:
The prototype looks good I have a couple of comments:
 - The section titles are not all placed the same way: Source of mapping suggestion and Source of ontology term information are in line with the logo but Trait name sources is above. Also the first two are lowercase and the third upper case. I would suggest they should all look like Trait name sources
 - I would place the Configure button on the right of the import now button
 - The configure expand arrow should pointing to the right  when collapsed and down when expanded

joj0s:
> The section titles are not all placed the same way

Regarding this one, the idea is that 'TRAIT NAME SOURCES` is supposed to be a general section title, meaning that more trait name sources can be added later, besides ClinVar, as a list, beneath the 'TRAIT NAME SOURCES' section. I'm not sure if that makes sense? If not, let me know so I can clarify more on that. 

[...]

PR #39, opened 2020-06-16
ClinVar data fetching and parsing

### LINKED ISSUE #14: Implement import of trait names from ClinVar

joj0s:
This PR adds a script to download and parse data from ClinVar, as well as a dummy 'datasources' page, containing a button to trigger this proccess. Closes #14


traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+from ..models import Trait
+
+
+def extract_clinvar_data():
+    """
+    This function downloads the latest ClinVar TSV release data and extracts it into a 
+    'variant_summary.txt' file
+    """
+    print("Downloading ClinVar data...")
+    url = 'https://ftp.ncbi.nlm.nih.gov/pub/clinvar/tab_delimited/variant_summary.txt.gz'
```
tskir, line 16: I think it's better to define this as a constant at the top of the module. Eventually this will be a configurable parameter (see https://github.com/EBIvariation/trait-curation/issues/32).

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+    urllib.request.urlretrieve(url, './variant_summary.txt.gz')
+    print("Extracting ClinVar data...")
+    with gzip.open('variant_summary.txt.gz', 'rb') as f_in:
+        with open('variant_summary.txt', 'wb') as f_out:
+            shutil.copyfileobj(f_in, f_out)
```
tskir, line 21: I wonder if explicitly unpacking this file is necessary—presumably, inside `parse_trait_names_and_source_records` we could just call `gzip.open` directly on the gzipped file instead of `open`.

(However, downloading the file should be still left in a separate function.)

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+import shutil
+import itertools
+import os
+from ..models import Trait
+
+
+def extract_clinvar_data():
+    """
+    This function downloads the latest ClinVar TSV release data and extracts it into a 
+    'variant_summary.txt' file
```
tskir, line 13:
```suggestion
    This function downloads the latest ClinVar TSV release data and extracts it into a 'variant_summary.txt' file.
```

Since this fragment is 114 characters, which is less than 120, it should fit into one line. Obviously this is not a huge issue, but it would be better to have consistent line lengths. I'm not familiar with VSCode, but presumably it should have a feature to display a vertical line after a (configurable) number of columns, to demonstrate where lines should be wrapped.

To demonstrate what I'm talking about, this is how this feature looks in PyCharm (screenshot is not mine, some random pic from the web):
image

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+    with gzip.open('variant_summary.txt.gz', 'rb') as f_in:
+        with open('variant_summary.txt', 'wb') as f_out:
+            shutil.copyfileobj(f_in, f_out)
+
+
+def parse_trait_names_and_source_records():
+    """
+    This function parses a downloaded 'variant_summary.txt' file, and returns a unique set of traint names
+    along with their calculated source record number, in a form of a dictionary where the key is the trait name
+    and the value is the source record number
```
tskir, line 28:
```suggestion
    This function parses a downloaded 'variant_summary.txt' file, and returns a unique set of trait names
    along with their calculated source record number, in a form of a dictionary where the key is the trait name
    and the value is the source record number.
```

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    print("Parsing ClinVar data...")
+    with open('variant_summary.txt') as tsvfile:
+        reader = csv.DictReader(tsvfile, dialect='excel-tab')
+        # A dictionary with trait names as keys and sets of source records as values
+        traits_dict = dict()
+        # The int value here defines how many records should be parsed
+        for row in itertools.islice(reader, 10):
```
tskir, line 36: Ah, I was just wondering why only 10 records get imported :) This is absolutely fine—we don't want to import everything, which would take forever and probably overrun free Heroku DB limits.

However, could you please do two things:
* Define this as a constant at the top of the module (in the future this will also be a user-configurable parameter, see #32)
* Raise the limit from 10 to, say, 100, so that the selection of test traits is a bit larger

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+            # For every row, get its allele_id and all its rcv_accessions and phenotypes
+            row_alleleid = [(row['#AlleleID'])]
+            row_rcv_list = set(row['RCVaccession'].split(';'))
+            row_phenotype_list = set(row['PhenotypeList'].split(';'))
+            # Get every possible combination tuple of allele_id rcv_accessions and phenotypes for the current row
+            tuple_list = list(itertools.product(
+                row_alleleid, row_rcv_list, row_phenotype_list))
```
tskir, line 43: Looks like that this part is currently not how it's supposed to be. Consider the example:
* **row_alleleid** = [42]
* **row_rcv_list** = [RCV100, RCV200]
* **row_phenotype_list** = [Disease A, Disease B]

If I understand correctly, itertools.product will produce four lists:
* [42, RCV100, Disease A]
* [42, RCV100, Disease B]
* [42, RCV200, Disease A]
* [42, RCV200, Disease B]

However, items in **row_rcv_list** and **row_phenotype_list** cannot be conbined freely, they correspond pairwise to *each other.* So in this case, the output should be:
* [42, RCV100, Disease A]
* [42, RCV200, Disease B]

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+        # A dictionary with trait names as keys and sets of source records as values
+        traits_dict = dict()
+        # The int value here defines how many records should be parsed
+        for row in itertools.islice(reader, 10):
+            # For every row, get its allele_id and all its rcv_accessions and phenotypes
+            row_alleleid = [(row['#AlleleID'])]
+            row_rcv_list = set(row['RCVaccession'].split(';'))
+            row_phenotype_list = set(row['PhenotypeList'].split(';'))
+            # Get every possible combination tuple of allele_id rcv_accessions and phenotypes for the current row
+            tuple_list = list(itertools.product(
```
tskir, line 42: I think `list(...)` wrap is redundant here. itertools.product returns a generator which can and should be iterated directly.

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+            traits_dict[key] = len(traits_dict[key])
+        os.remove("variant_summary.txt")
+        os.remove("variant_summary.txt.gz")
+        return traits_dict
+
+
+def store_data(traits_dict):
+    """
+    This function accepts a dictionary in the form of keys=trait names and values=source record numbers
+    and stores them in the database using the Django ORM
```
tskir, line 62:
```suggestion
    and stores them in the database using the Django ORM.
```

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    print("Storing ClinVar data...")
+    try:
```
tskir, line 65: Since errors could also happen during download and parse stages, I'm thinking a better place to handle exceptions would be inside fetch_data()

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+        for trait_name in traits_dict.keys():
+            print(trait_name)
+            if Trait.objects.filter(name=trait_name).exists():
+                trait = Trait.objects.filter(name=trait_name).first()
+                trait.number_of_source_records = traits_dict[trait_name]
+                trait.save()
+            else:
+                trait = Trait(name=trait_name, status="unmapped",
+                              number_of_source_records=traits_dict[trait_name])
+                trait.save()
```
tskir, line 75: This is not a review comment, but I just wanted to say that elegant constructs like this is why I like Python. Recently I had to work on an existing project which is written in Java and uses Spring ORM. This is how [a similar operation](https://github.com/EBIvariation/metadata-ws/blob/master/metadata-ws/src/main/java/uk/ac/ebi/ampt2d/metadata/persistence/repositories/FileRepositoryCustom.java) (find-or-save) looks like in that world. That stuff gave me nightmares.

traitcuration/traits/templates/traits/datasources.html
```diff
@@ -0,0 +1,16 @@
...
+  <a href="{% url 'fetch_data' %}"
+    ><button class="uk-button uk-button-primary">FETCH DATA</button>
+  </a>
+  {% if messages %}
+    <ul class="messages">
+        {% for message in messages %}
+        <li {% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
+        {% endfor %}
+    </ul>
+    {% endif %}
```
tskir, line 14: Maybe I'm confused, but what is the `<ul class="messages">` block supposed to do? I don't see message.tags populated anywhere, and the `{% if messages %}` conditional never seems to fire.

tskir:
@joj0s Also some more comments:
* There appear to be minor merge conflicts after merging two previous PRs
* You might want to include changes from https://github.com/EBIvariation/trait-curation/issues/40 which are follow-ups from PR #35, although you could also do that in a separate PR if you'd like

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
+import urllib.request
+import csv
+import gzip
+import shutil
+import itertools
+import os
+from ..models import Trait
```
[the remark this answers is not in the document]
joj0s: Yeah, I was hoping the flake8 linter would point some of those details out but it turns out it doesn't, so please do keep correcting me on those. In the meantime I will read the full PEP8 manual though

traitcuration/traits/datasources/clinvar.py
joj0s: Thanks for the suggestion, I did something similar where I added a ruler in the editor to show the 120 character breakpoint

traitcuration/traits/datasources/clinvar.py
joj0s: Yes, since I had the source record calculation wrong, I removed this line altogether

traitcuration/traits/templates/traits/datasources.html
joj0s: Yes, that's my bad, i was testing Django messages and forgot to remove that from the page

traitcuration/traits/datasources/clinvar.py
joj0s: I am sadly familiar, I applied for https://github.com/EBIvariation/eva-ws last year :'(

traitcuration/traits/datasources/clinvar.py
joj0s: You are right, I moved the exception handling to the view

traitcuration/traits/datasources/clinvar.py
joj0s: I did that and I raised the limit to 200

traitcuration/traits/datasources/clinvar.py
joj0s: You are right, I removed the extraction and added opening via gzip

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
+"""
+This module contains utility functions for downloading, parsing and storing trait data fron ClinVar
+"""
+
+import urllib.request
+import csv
+import gzip
+import itertools
+import os
```
tskir, line 9:
```suggestion
import csv
import gzip
import itertools
import os
import urllib.request
```

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+        for row in itertools.islice(reader, NUMBER_OF_RECORDS):
+            # For every row, get its allele_id and all its rcv_accessions and phenotypes
+            row_alleleid = (row['#AlleleID'])
+            row_rcv_list = row['RCVaccession'].split(';')
+            row_phenotype_list = row['PhenotypeList'].split(';')
+            # Get every possible pair tuple of allele_id rcv_accessions and phenotypes for the current row
+            tuple_set = set()
+            for index in range(len(row_rcv_list)):
+                tuple_set.add(
+                    (row_alleleid, row_rcv_list[index], row_phenotype_list[index]))
```
tskir, line 48: This can be potentially simplified with a set comprehension expression + `zip` function:

```suggestion
            # Get every possible pair tuple of allele_id rcv_accessions and phenotypes for the current row
            tuple_set = {(row_alleleid, rcv, phenotype) for rcv, phenotype in zip(row_rcv_list, row_phenotype_list)}
```

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+            # Insert the tuple in the dictionary
+            for tuple in tuple_set:
+                trait_name = tuple[2]
+                if trait_name in traits_dict:
+                    traits_dict[trait_name].update([tuple])
+                else:
+                    traits_dict[trait_name] = {tuple}
```
tskir, line 55: This can be simplified a little bit by using `setdefault` method of a dictionary:
```suggestion
                traits_dict.setdefault(trait_name, set())
                traits_dict[trait_name].update([tuple])
```

Actually, given that sets are mutable objects, one could even write it as a one-liner:

```python
traits_dict.setdefault(trait_name, set()).update([tuple])
```

But while this is neat, it also looks cryptic, so I try to avoid this.

traitcuration/traits/datasources/clinvar.py
```diff
@@ -0,0 +1,77 @@
...
+    print("Storing ClinVar data...")
+    for trait_name in traits_dict.keys():
+        if Trait.objects.filter(name=trait_name).exists():
+            trait = Trait.objects.filter(name=trait_name).first()
+            trait.number_of_source_records = traits_dict[trait_name]
+            trait.save()
+        else:
+            trait = Trait(name=trait_name, status="unmapped",
+                          number_of_source_records=traits_dict[trait_name])
+            trait.save()
```
tskir, line 77: Looks like `trait.save()` can be moved from the if/else as a single final command:

```suggestion
        if Trait.objects.filter(name=trait_name).exists():
            trait = Trait.objects.filter(name=trait_name).first()
            trait.number_of_source_records = traits_dict[trait_name]
        else:
            trait = Trait(name=trait_name, status="unmapped",
                          number_of_source_records=traits_dict[trait_name])
        trait.save()
```

Issue #40, opened 2020-06-17
Small follow-up actions from code reviews


tskir:
This meta-issue will track various small follow-up actions from code review iterations.

- [x] https://github.com/EBIvariation/trait-curation/pull/35#discussion_r440161830 Reformat two lines into a single line
- [x] https://github.com/EBIvariation/trait-curation/pull/35#discussion_r441881504 Avoid code duplication in the list of statuses by moving some of them into a variable
- [x] https://github.com/EBIvariation/trait-curation/pull/35#discussion_r440158511 Use rem units globally in the CSS

tskir:
They've been fixed by #45 so I'm going to close this issue. If we have small follow-up actions in the future, we can create new issues

PR #43, opened 2020-06-18
Trait sidebar and main trait information

### LINKED ISSUE #21: Display main information on a single trait page
image

While the UI mockup provides a general idea on how to organise the blocks, also keep in mind that trait names can be quite long. The longest which is currently present in the ClinVar database is "Capillary malformation of the lower lip, lymphatic malformation of face and neck, asymmetry of face and limbs, and partial/generalized overgrowth", at 145 characters long. We should ensure that the layout is at least not _badly_ broken by such trait names.

### LINKED ISSUE #24: Implement the side bar
image

In this issue, only the side bar itself is implemented and the “Review” button will not be functional yet.

Also, it looks like the trait status ("Under review" in the mockup example page) is present twice on this page, which is probably not necessary.

joj0s:
This PR adds a sidebar which shows review information and the timestamp of a trait's last update, as per the Figma prototype. It also adds clear display of a trait's main information

Closes #21, closes #24

### DIFF
--- a/traitcuration/static/styles/layout/layout.scss
+++ b/traitcuration/static/styles/layout/layout.scss
@@ -2,3 +2,4 @@
 
 @import "./footer/footer.scss";
 @import "./header/header.scss";
+@import "./sidebar/sidebar"
--- a/traitcuration/static/styles/layout/layout.scss
+++ b/traitcuration/static/styles/layout/layout.scss
@@ -1,5 +1,5 @@
 /* Styling for layout components */
 
-@import "./footer/footer.scss";
-@import "./header/header.scss";
+@import "./footer/footer";
+@import "./header/header";
 @import "./sidebar/sidebar"
--- a/traitcuration/traits/templates/traits/trait_detail.html
+++ b/traitcuration/traits/templates/traits/trait_detail.html
@@ -5,6 +5,37 @@ Trait Detail
 {% endblock %}
 
 {% block content %}
+<aside class="sidebar">
+    <div class="sidebar__wrapper">
+        <div class="sidebar__section">
+            <p class="sidebar__section-title">LAST UPDATED</p>
+            <span class="sidebar__section-text"> {{trait.timestamp_updated}} </span>
+            <hr/>
+        </div>
+        <div class="sidebar__section">
+            {% if trait.current_mapping %}
+            <p class="sidebar__section-title">CURATOR</p>
+            <span uk-icon="user" class="uk-margin-small-right"></span>
+            <span class="sidebar__section-text"> {{trait.current_mapping.curator}} </span>
+            <hr/>
+            {% endif %}
+        </div>
+        <div class="sidebar__section">
+            {% if trait.current_mapping %}
+            <p class="sidebar__section-title">REVIEW STATUS</p>
+            {% if trait.current_mapping.is_reviewed %}
+            <span class="uk-label label--success"> REVIEWED</span>
+            {% else %}
+            <span class="uk-label label--warning"> IN REVIEW </span>
+            {% endif %}
+            <button class="uk-button button--success sidebar__section-review-button"> 
+                <span uk-icon="check"> REVIEW </span> 
+            </button>
+            {% endif %}
+        </div>
+    </div>
+</aside>
+
 <div class="uk-container uk-container-xlarge">
     <h1>Trait details</h1>
     <ul class="uk-list">
--- a/traitcuration/traits/templates/traits/trait_detail.html
+++ b/traitcuration/traits/templates/traits/trait_detail.html
@@ -4,6 +4,7 @@
 Trait Detail
 {% endblock %}
 
+{% load trait_filters %}
 {% block content %}
 <aside class="sidebar">
     <div class="sidebar__wrapper">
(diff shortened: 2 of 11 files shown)


traitcuration/static/styles/layout/layout.scss
```diff
@@ -2,3 +2,4 @@
 
 @import "./footer/footer.scss";
 @import "./header/header.scss";
+@import "./sidebar/sidebar"
```
tskir, line 5: Is there a reason why footer and header are imported using their filenames directly (with `.scss` extension), but sidebar in some different way? I'm not an expert in Sass, so the reason is probably there and I just don't know it

traitcuration/traits/templates/traits/trait_detail.html
```diff
@@ -1,26 +1,82 @@
...
+      <span class="info-section__value">{{trait.name}}</span>
+    </div>
+    <div class="info-section__column">
+      <span class="info-section__title">SOURCE RECORDS</span>
+      <span class="info-section__value"
+        >{% firstof trait.number_of_source_records '0' %}</span
+      >
+    </div>
+    <div class="info-section__column">
+      <span class="info-section__title">TRAIT STATUS</span>
```
tskir, line 65: When a trait name is long but window width is not, something like this happens:

image

In this case it would be better to make "trait name" column smaller, and to ensure that "source records" and "trait status" will always fit

traitcuration/traits/templates/traits/trait_detail.html
```diff
@@ -1,26 +1,82 @@
...
+      </button>
+      {% endif %}
+    </div>
+  </div>
+</aside>
+
+<div class="uk-container uk-container-xlarge trait-container">
+  <div class="info-section__wrapper">
+    <div class="info-section__column">
+      <span class="info-section__title">TRAIT NAME</span>
```
tskir, line 55: Currently, the contents of the trait page are indented compared to the "Browse traits" button in the header:

image

I'm thinking it would be better to align them all to the same left margin

traitcuration/static/styles/layout/layout.scss
joj0s: Actually no, the files with `.scss` extension were part of the scss folder structure template I am using and the sidebar was imported manually, that's the cause of the difference. I will define a follow up action to do it one way across all imports to maintain consistency

traitcuration/traits/templates/traits/trait_detail.html
joj0s: Well, the norm on headers is for them to have a smaller indent compared to the main page containers (you can look at GitHub here as an example). So my suggestion here would be to actually decrease the indent by a small amount on the header so that the distinction between itself and the container is clearer. 

Another reason for that is that in very large screens, the container will have more margin compared to smaller screens so basically the only way to align those two is to put the header links inside the same type of container

traitcuration/traits/templates/traits/trait_detail.html
joj0s: Yes I will do that. Another thing for really small screens (basically just tablets in portrait mode), I could make it so that the trait status wraps beneath the first row to save some space. What do you think?

traitcuration/static/styles/layout/layout.scss
tskir: OK! In this case, could you either fix it in this PR, or add a follow-up action in existing ticket https://github.com/EBIvariation/trait-curation/issues/40, or create a new ticket for this action? Just so that it doesn't get lost in the discussions

traitcuration/traits/templates/traits/trait_detail.html
tskir: I'm afraid it may mess with the layout and how the users expect the screen do behave when resizing its window. A trait status is an important marker, and people will expect to see it always in the same place. I think that at this point just changing column margins and re-wrapping trait name into multiple rows should be enough, especially considering that few trait names will be this long.

traitcuration/traits/templates/traits/trait_detail.html
tskir: Right, so first of all, one the second thought I do agree with you about the indents. Content having larger intent than header makes sense.

However, the problem which I noticed now is that margins changes drastically on different zoom levels:

100%
image

125%
image

150%
image

I think the look closest to normal is on 125% level. But it should look approximately the same at least in the 100–150% zoom range, as those are frequently used.

This can be done either in this PR, or as a folow-up ticket (perhaps united with comment from [here](https://github.com/EBIvariation/trait-curation/pull/43#discussion_r446364192))

joj0s: That's weird. I am using the `container` class that comes with UIKit but I might have to create one my own. Could you tell me the width of the screen you tested this on?

traitcuration/traits/templates/traits/trait_detail.html
joj0s: I agree, I will add that

joj0s: So, I did fix it by making the sidebar take up 14% of the page which ensures that the trait column always wraps, and it looks like this
1. On a 1920x1080 screen
image
2. On a 1360x768px screen
image
3. On a 1200x800px tablet
image

As you can imagine the page will be pretty unusable on anything smaller than a landscape tablet. So I could add some media queries to do something like increase the sidebar width and decrease the font size in screens smaller than 800px if that is a concern. Of course I am not referring to having it work on really small screens such as mobile phones, because that would require major redesigning as we discussed.

traitcuration/static/styles/layout/layout.scss
joj0s: Yes, I am adding that in this PR

traitcuration/traits/templates/traits/trait_detail.html
tskir: I can confirm that what we have currently is acceptable for reasonable window width and zoom range of 100-150%, so we may consider this issue resolved. Thanks!

traitcuration/traits/templates/traits/trait_detail.html
tskir: I don't know what changed (or whether anything changed at all), but now I don't see this problem and zoom works as expected. Marking this as resolved then :)

traitcuration/static/styles/layout/layout.scss
tskir: Great, this means that this action is the only one remaining for this PR. Once you address it, feel free to merge this PR without further review from my side

joj0s:
@tskir I actually pushed fixes this morning both for the `.scss` extension issue as well as the trait name wrapping. It seems the push was unsuccessful due to GitHub's techinical problems. So I will push them again and it would be better to check again if the layout is as expected

tskir:
@joj0s I have checked code & deployment again. Both the files and the layout in the app are fine, I'm going to go ahead and merge this

PR #45, opened 2020-06-19
Follow up actions

joj0s:
This PR addresses the current open follow-up actions as described in #40.

Regarding px units, I only kept them in box shadows and single px borders. It has PR #43 as its base branch so that I can change px values there as well, so I suggest we merge it after PR #43


tskir:
This doesn't have merge conflicts so I'm going to go ahead and merge this one as well

PR #49, opened 2020-06-29
Query ZOOMA to get possible mappings

### LINKED ISSUE #30: Implement querying ZOOMA to get possible mappings for a trait
This should essentially be done by copying the code from the main Open Targets repository, specifically from the [`trait_mapping`](https://github.com/EBIvariation/eva-cttv-pipeline/tree/master/eva_cttv_pipeline/trait_mapping) directory.

joj0s:
This PR adds the ability to get mapping suggestions from ZOOMA and create terms and suggested mappings in the app's database accordingly. It also adds basic OLS queries to get info about the retrieved term. Closes #30


joj0s:
One thing I'd like to discuss about this PR is this. For imported non-EFO terms, I am saving their original URIs in our database. For example, for MONDO:0001933, I am saving http://purl.obolibrary.org/obo/MONDO_0001933. In order to get the EFO link, I would have to attach that to an EFO prefix, such as https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FMONDO_0001933. So which of these would we want to save?

traitcuration/traits/datasources/ols.py
```diff
@@ -0,0 +1,36 @@
+import requests
```
tskir, line 1: Could you add a (simple) docstring please?

traitcuration/traits/datasources/ols.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+BASE_URL = "https://www.ebi.ac.uk/ols/api/"
+
+
+def make_ols_query(term_uri, ontology):
+    """
+    Takes in a term URI (or IRI as referenced in OLS documentation) and the ontology id to search against. Returns a
```
tskir, line 9: I've finally taken some time to research the different abbreviations. Looks like IRI is the most inclusive term, which is basically a URI which can have _any_ characters in it, not just ASCI, and it does not have to have percent encoding.

So I suggest we change all occurrences of "uri" to "iri", and it would also help to include a short description of what a IRI is.

tskir: Essentially (if I've got that right),

IRI = https://en.wiktionary.org/wiki/tréma
URI = https://en.wiktionary.org/wiki/tr%C3%A9ma (**must** be percent encoded)

traitcuration/traits/datasources/ols.py
```diff
@@ -0,0 +1,36 @@
...
+    dictionary for that term with the fields 'label', 'obo_id' which is the CURIE and 'status'.
+    """
+    results = requests.get(
+        f"{BASE_URL}ontologies/{ontology}/terms?iri={term_uri}").json()
```
tskir, line 13:
```suggestion
    results = requests.get("{BASE_URL}ontologies/{ontology}/terms?iri={term_uri}").json()
```

Looks like this can fit easily into <120 character line. Is it at all possible that a linter you use for Python misbehaves and has its line limit set to the default 80, producing these things from time to time?

traitcuration/traits/datasources/ols.py
```diff
@@ -0,0 +1,36 @@
...
+    print(f"{BASE_URL}ontologies/{ontology}/terms?iri={term_uri}")
+    if 'error' in results:
+        return None
```
tskir, line 16: Do you know in which cases it returns an 'error'? Is it only for missing terms, or can there be other scenarios?

traitcuration/traits/datasources/ols.py
```diff
@@ -0,0 +1,36 @@
...
+    term_info = results["_embedded"]["terms"][0]
+    term_curie = term_info["obo_id"]
+    term_label = term_info["label"]
```
tskir, line 19: When parsing a response, it's useful to provide example values in comments to make it more readable. E.g.:

```python
term_curie = term_info["obo_id"]  # E.g. 'MONDO:123456'
term_label = term_info["label"]  # E.g. 'Some disease name'
```

traitcuration/traits/datasources/ols.py
```diff
@@ -0,0 +1,36 @@
...
+    print(info_dict)
+    print()
+    return info_dict
+
+
+def get_ontology_id(term_uri):
+    """
+    Extracts the ontology id from the term uri, to be used for OLS queries
+    """
+    ontology_id = term_uri.split('/')[-1].split('_')[0].lower()
```
tskir, line 33: Could you provide an example of input/output values to make it more clear which transformations are happening?

traitcuration/traits/datasources/ols.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    results = requests.get(
+        f"{BASE_URL}ontologies/{ontology}/terms?iri={term_uri}").json()
+    print(f"{BASE_URL}ontologies/{ontology}/terms?iri={term_uri}")
+    if 'error' in results:
+        return None
+    term_info = results["_embedded"]["terms"][0]
+    term_curie = term_info["obo_id"]
+    term_label = term_info["label"]
+    term_status = 'current' if ontology == 'efo' else 'needs_import'
```
tskir, line 20: It looks to me that this is not the best place to decide the final term status. I get the logic (you first call EFO, and other ontologies are queried _only if_ nothing was found in EFO), but it's better to make it separate and explicit. That is, `make_ols_query` should only return whether the requested term is in the requested ontology; and the final trait status calculations should be done outside of it in `zooma.py`

traitcuration/traits/datasources/zooma.py
```diff
@@ -0,0 +1,64 @@
+import requests
```
tskir, line 1: And a small docstring here as well, please

traitcuration/traits/datasources/zooma.py
```diff
@@ -0,0 +1,64 @@
...
+
+from ..models import Trait, MappingSuggestion, OntologyTerm, User
+from .ols import make_ols_query, get_ontology_id
+
+
+BASE_URL = "http://www.ebi.ac.uk/spot/zooma/v2/api"
```
tskir, line 7: I'm wondering if we should start collecting all base URLs and configuration settings in a separate module (`config`?) to make it easier in the future to make all of them customizable via a configuration page.

This is not necessary to do as part of this PR, just something to consider

traitcuration/traits/datasources/zooma.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Retrieves zooma mapping suggestions for all traits and creates terms and suggestions in the app's database for each
+    of them.
+    """
+    traits = Trait.objects.all()
+    for trait in traits:
+        results_list = get_suggestions_for_trait(trait)
+        for result in results_list:
+            term = create_local_term(result)
+            if term is not None:
```
tskir, line 20: I think there should be a comment here explaining in which situations a term can be None, and what this implies

traitcuration/traits/datasources/zooma.py
```diff
@@ -0,0 +1,64 @@
...
+    term_info = make_ols_query(term_uri, 'efo')
+    if term_info is None:
+        term_info = make_ols_query(term_uri, get_ontology_id(term_uri))
+        # If the term is not found in the original ontology either
+        if term_info is None:
+            print(f'No info found on {term_uri}')
+            return
+    # Create an ontology term in the database
+    term = OntologyTerm(
+        curie=term_info['curie'], uri=term_uri, label=term_info['label'], status=term_info['status'])
```
tskir, line 53:
```suggestion
    term = OntologyTerm(curie=term_info['curie'], uri=term_uri, label=term_info['label'], status=term_info['status'])
```

Fits into one line <120 characters

tskir:
(Also there's now a merge conflict)

traitcuration/traits/datasources/ols.py
joj0s: Sure, I'll change that in the data model as well

traitcuration/traits/datasources/ols.py
joj0s: Yeah I have given that some thought. The only cases for errors I have found are '404 not found', the most common one for missing terms, and network error. Should I handle network errors any differently?

traitcuration/traits/datasources/zooma.py
joj0s: Hmm yes that seems like a good option, I will create an issue for that.

traitcuration/traits/datasources/ols.py
joj0s: Yes, this is caused by autopep8. I thought I had fixed that, but now it seems it's fixed for good, sorry for the inconvenience

joj0s:
@tskir I rebased against master but there were conflicting migrations. Should I merge the migrations and commit them here?

tskir:
> @tskir I rebased against master but there were conflicting migrations. Should I merge the migrations and commit them here?

I've never actually worked with Django migrations myself and I don't know which is the accepted practice in such circumstances. On the surface, it looks like there are two ways here:
* Merge two 0009* migrations into one manually
* Delete the more recent 0009* migration and re-generate it again, in which case presumably it will get prefix 0010* and will depend on 0009_auto_20200619_0809

joj0s:
In these cases, Django recommends merging them. The merge created a 0010* migration that depends on the two 0090* ones. I will push them and it should work fine on the deployment

joj0s:
I can confirm everything looks good regarding the migrations, and the Heroku deployment is running fine.

tskir:
> In these cases, Django recommends merging them.

Huh, this is interesting. I mean, whichever works, and for this simple case it certainly doesn't matter at all. But intuitively it looks to me that migrations dependency graph should be kept linear whenever possible, precisely so as not to deal with any possible conflicts. Again, for this simple case either solution is certainly fine, let's use the merge.

> I can confirm everything looks good regarding the migrations, and the Heroku deployment is running fine.

Great! I'll take a look at this later today

traitcuration/traits/datasources/ols.py
tskir: Yes, network errors definitely have to be handled separately, as those "errors" are two really different cases:

- Term is (really) not found in EFO → requires import
- Term is "not found" in EFO due to network error → request should be retried, otherwise we will either get an incorrect status (when querying OLS) or we will miss some suggestions (when querying ZOOMA)

I suggest you handle requests in the following manner:
* For network errors, catch-and-retry with `@retry` decorator (see example [here](https://github.com/EBIvariation/vep-mapping-pipeline/blob/eee47471312cbcf08955acdd68d76851c5ffa5b4/vep_mapping_pipeline/consequence_mapping.py#L54)). If for a given entity we have exceeded the maximum number of retries, the entire update process should be halted and an error raised.
  + By the way, in production, all update operations (from ClinVar or from ZOOMA) will have to be started as transactions and rolled back in case of any exception, with the error being logged. I have created a future issue for this: https://github.com/EBIvariation/trait-curation/issues/52
* For _expected_ HTTP error codes, like a term being not found in EFO, verify that the error is of the right type: check the code and possibly also the contents of the response body (if applicable)
* For all other _unexpected conditions_ (like HTTP error codes which are not 404) raise a log an exception

traitcuration/traits/datasources/ols.py
```diff
@@ -0,0 +1,41 @@
...
+    if 'error' in results:
+        print
+        return None
+    term_info = results["_embedded"]["terms"][0]
+    term_curie = term_info["obo_id"]  # E.g. EFO:0000400
+    term_label = term_info["label"]  # E.g. Diabetes mellitus
+    term_is_obsolete = term_info["is_obsolete"]  # True or False value on whether the term is obsolete or not
+    info_dict = {"curie": term_curie, "label": term_label, "is_obsolete": term_is_obsolete}
+    print(info_dict)
+    print()
```
tskir, line 27: Not a necessary change (at least for now) but a suggestion: given multiple logging commands (some of which will make it into production, some required only for development), it would be slightly easier to use the `logging` module and to define specific calls as "error", "info", "debug" etc.

Example of using the logging module: https://github.com/EBIvariation/vep-mapping-pipeline/blob/master/vep_mapping_pipeline/consequence_mapping.py (Ctrl+F for "logging" and then "logger")

traitcuration/traits/datasources/zooma.py
```diff
@@ -0,0 +1,84 @@
...
+    of them.
+    """
+    traits = Trait.objects.all()
+    for trait in traits:
+        results_list = get_suggestions_for_trait(trait)
+        for result in results_list:
+            term = create_local_term(result)
+            # If local term creation returns None, it means that this term already exists in the database
+            if term is not None:
+                create_mapping_suggestion(trait, term)
```
tskir, line 26: Looks like this only adds suggestions but never removes the old ones. I think during the suggestion update cycle we can also remove old suggestions for a trait which are:
* Originally created by ZOOMA
* **Not** selected as an accepted suggestion for this trait

tskir: (Otherwise the list of suggestions will just grow and grow and grow)

traitcuration/traits/datasources/zooma.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    Retrieves zooma mapping suggestions for all traits and creates terms and suggestions in the app's database for each
+    of them.
+    """
+    traits = Trait.objects.all()
+    for trait in traits:
+        results_list = get_suggestions_for_trait(trait)
+        for result in results_list:
+            term = create_local_term(result)
+            # If local term creation returns None, it means that this term already exists in the database
```
tskir, line 24: Not sure if I get the logic here. What about the case when the term _is_ already in the database? Looks like `if term is not None` will be False and the mapping suggestion will not be created. But this shouldn't be the case. Consider the example:
* Originally "Trait A" had a ZOOMA suggestion to TERM:123456
* During next update iteration, we ingest "Trait A, some subtype", and ZOOMA also suggests to map it to TERM:123456. In this case, we should just create the mapping suggestion linking "Trait A, some subtype" to the _existing_ term.

tskir: This can be most easily implemented by returning the existing term from the database (instead of None) in `create_local_term`

traitcuration/traits/datasources/zooma.py
```diff
@@ -0,0 +1,84 @@
...
+    if OntologyTerm.objects.filter(iri=term_iri).exists():
+        return
+    # Search for the term in EFO and return its information. If it is not in EFO, search for it in original ontology.
+    term_info = make_ols_query(term_iri, 'efo')
+    term_ontology_id = 'efo'
+    # None here means that the term wasn't found in the EFO ontology
+    if term_info is None:
+        term_ontology_id = get_ontology_id(term_iri)
+        term_info = make_ols_query(term_iri, term_ontology_id)
+        # If the term is not found in the original ontology either, return
```
tskir, line 55: This situation (which is really unusual) should immediately change the term status to "deleted" (we do have "deleted" in the final db model, right?)

traitcuration/traits/datasources/zooma.py
```diff
@@ -0,0 +1,84 @@
...
+    return term
+
+
+def get_term_status(ontology_id, is_obsolete):
+    """
+    Takes the ontology_id of a term and a whether it is obsolete or not, and returns its calculated status.
+    'Obsolete' if the is_obsolete flag is true, 'Current' if its ontology is EFO, and 'Needs Import' otherwise
+    """
+    if is_obsolete:
+        return 'obsolete'
```
tskir, line 72: This should ideally use not the literal "obsolete", but some enum value which is defined just once for the entire project. I think we already have one somewhere in data models

traitcuration/traits/datasources/zooma.py
```diff
@@ -0,0 +1,84 @@
...
+    return 'needs_import'
+
+
+def create_mapping_suggestion(trait, term):
+    """
+    Creates a mapping suggestion in the app's database.
+    """
+    zooma = User.objects.filter(username="Zooma").first()
+    suggestion = MappingSuggestion(trait_id=trait, term_id=term, made_by=zooma)
+    suggestion.save()
```
tskir, line 84: What happens if the mapping suggestion for this combination of trait and term already exists? Looks like it is going to get duplicated.

BTW, is there a way to specify complex UNIQUE constraints in the Django data model? Like, (trait_id, term_id) pair should be unique across all mapping suggestions and this should be enforced.

traitcuration/traits/datasources/ols.py
joj0s: Regarding the error handling, just to clarify that I am proceeding as I should, I am first checking if I get a 404, in which case I return `None` as the result of the query. In any other case, I am using `response.raise_for_status()` to raise an Exception. Should I retry the query in case of unexpected HTTP errors or just for network errors?

traitcuration/traits/datasources/zooma.py
joj0s: So basically remove any preexisting suggestions for a trait made by ZOOMA and repopulate them in each cycle, right?

traitcuration/traits/datasources/zooma.py
joj0s: https://docs.djangoproject.com/en/3.0/ref/models/options/#unique-together This is what we need I think, I will add that

traitcuration/traits/datasources/zooma.py
joj0s: OK so the enum situation is kind of confusing right now. This is what we currently have:

- A list of possible status fields (values and human-readable labels) for traits and terms *outside* of the model classes, so that it can be shared between those two, and just have the trait status field add another value to this array, for the 'Awaiting Review' status.
```
STATUS_FIELDS = [
    ('current', 'Current'),
    ('unmapped', 'Unmapped'),
    ('obsolete', 'Obsolete'),
    ('deleted', 'Deleted'),
    ('needs_import', 'Needs Import'),
    ('awaiting_import', 'Awaiting Import'),
    ('needs_creation', 'Needs Creation'),
    ('awaiting_creation', 'Awaiting Creation'),
]
```

- Models using these choices as such:
 ```
class Trait(models.Model):
    status = models.CharField(max_length=50, choices=STATUS_FIELDS + [('awaiting_review', 'Awaiting Review')])
    
class OntologyTerm(models.Model):
    status = models.CharField(max_length=50, choices=STATUS_FIELDS)
```

This provides enum functionality and validation if a value which is not in STATUS_FIELDS is provided. 

The problem with that if we don't want to use literals to insert values, especially since the STATUS_FIELDS array is defined outside of each class, is that getting the enum value defined in the models is kind of hard to read and hard to use. For example, to get all possible values one would have to do `Trait._meta.get_field('status').choices` which would return the `STATUS_FIELDS` array exactly as defined, and then use array and tuple methods to get the value.

A possible alternative to that, as described here https://adamj.eu/tech/2020/01/27/moving-to-django-3-field-choices-enumeration-types/ is to create a TextChoices subclass for the status values:
```
from django.db import models


class Status(models.TextChoices):
    CURRENT = 'C', 'Current'
    UNMAPPED = 'U', 'Unmapped'


class Trait(models.Model):
    status = models.CharField(
        max_length=2,
        choices=Status.choices,
        default=Status.UNMAPPED,
    )
```

It can then be accessed like this `unmapped_traits = Trait.objects.filter(status=Status.UNMAPPED)`
The problem with that is that I'm not sure how we would add the 'awaiting_review' specifically to the Trait model, maybe another class that inherits?

traitcuration/traits/datasources/ols.py
tskir: > Regarding the error handling, just to clarify that I am proceeding as I should, I am first checking if I get a 404, in which case I return `None` as the result of the query.

Yes, this is fine, as 404 is an expected error in this case.

> In any other case, I am using `response.raise_for_status()` to raise potential exceptions. Should I let `retry` catch unexpected HTTP errors and make the query again (this is what is done in the example you provided) or should I have it catch only network errors?

Hmm. Yes, good question. On the one hand, everything but 200 or 404 is not expected from OLS. On the other hand, sometimes the server can have a hiccup and return an unusual code _sporadically,_ which probably shouldn't fail the entire process right away.

So yes, let's use the `retry` as specified in my example, where it catches both network errors from `requests.get()/post()` _and_ unusual sporadic HTTP responses from `raise_for_status()`. Thank you for noticing and asking about this!

traitcuration/traits/datasources/ols.py
tskir: I've replied in the thread which we already have for error handling above: https://github.com/EBIvariation/trait-curation/pull/49#discussion_r448694485. The GitHub interface is annoying in that, when some comment was made during previous review iterations, it will duplicate it for the newest iteration, but won't allow to actually comment

traitcuration/traits/datasources/zooma.py
tskir: Yes, correct. Also, thanks for noticing that the _previous_ mappings also shouldn't be removed, I didn't think of that at all.

Perhaps this removal could even be done by a single SQL command for all suggestions at once to improve performance (although this sort of optimisation is not necessary at this point).

traitcuration/traits/datasources/zooma.py
tskir: Yes, sorry, I didn't mean to imply that we would get the values _from_ the models, understandably this could be cumbersome. What I meant is that the `'current'` literal should only be defined at one place in the entire code, and then used by import from there.

The simplest solution I had in mind for that is converting the `STATUS_FIELDS` tuple into a plain [enum](https://docs.python.org/3/library/enum.html) and then, rather than using `'current'` inside zooma.py, use something like `STATUS_FIELDS.CURRENT.value`

The models.TextChoices class which you found looks very similar to what plain enums do. You can choose to use either it or enum, both seem fine to me.

Finally, regarding appending "awaiting review", I think a class which inherits from the main one is a good idea. I think enums support inheritance, and the TextChoices should certainly do since this is just a class

traitcuration/traits/datasources/zooma.py
joj0s: Yeah we do, but do we still want to create the term if we don't find it anywhere during OLS queries?

traitcuration/traits/datasources/zooma.py
joj0s: https://docs.python.org/3/library/enum.html#restricted-enum-subclassing Turns out ENUM subclassing is actually not possible

traitcuration/traits/datasources/zooma.py
tskir: Yes, I would say we still want to create this term. This situation means that either something is wrong on the ZOOMA side, or some bug happened on our side during OLS lookup. So it's better to have this more explicit (by showing up as a deleted term in a suggestion), so this could be noticed and fixed/reported.

traitcuration/traits/datasources/zooma.py
tskir: > Turns out enum field inheritance is actually not possible
Huh. Indeed. Didn't know that.

I'm really not happy with duplicating the core status values. There are two ways we can handle this:
* The author of the enum library suggests going with another library in such cases, aenum (advanced enumerations), which allows extensions and combinations. See his comments here: https://stackoverflow.com/a/33680929, https://stackoverflow.com/a/46080827.
* Alternatively, we could just have a single enum with statuses (including "awaiting_review"), and when defining a model, we could compute the list of allowable values based on that enum, taking all values _except_ the ones we don't want to be allowed.

joj0s: So the approach I've decided to use is mostly the second one you mentioned. I am defining a separate status class with all the possible statuses, and then two class methods, `trait_choices()` and `term_choices()` that filter the possible status values accordingly, so then the code in the Trait and Term models looks like this: 
`status = models.CharField(max_length=50, choices=Status.trait_choices())`

tskir: This sounds good! We would still be able to access individual statuses as well via some attributes, like `Status.CURRENT`, right? Since that was the whole point

joj0s: Yes yes, we will just need to import the Status class from `models.py` to do that

joj0s:
@tskir I have implemented the following changes:
- Added logging via `logging` module
- Added network error handling via `retry` decorator
- Enabled creation of 'deleted' terms if information about them is not found
- Merged possible Status fields into a single `TextChoices` class
- Added deletion of mapping suggestions that haven't been returned in the ZOOMA query and are not part of any previous mappings
- Added a `unique_together` constraint for mapping suggestions regarding term and trait combinations

PR #51, opened 2020-06-30
Dummy data import

### LINKED ISSUE #46: Implement dummy data importing for testing
Right now, when a new deployment is made to Heroku, a new database instance is created. It would be convenient to have a button to trigger import of dummy data so that testing of the app's UI and logic can be made easier

joj0s:
This is a PR on top of the Zooma workflow PR adds an easy way of importing dummy data that covers every possible trait status scenario, except for unmapped traits, which can be easily obtained through ClinVar. Closes #46

### DIFF
--- a/traitcuration/traits/datasources/dummy.py
+++ b/traitcuration/traits/datasources/dummy.py
@@ -1,7 +1,7 @@
 """
 This module holds dummy data to import during development
 """
-from ..models import OntologyTerm, MappingSuggestion, Trait, Mapping, User, Review
+from ..models import OntologyTerm, MappingSuggestion, Trait, Mapping, User, Review, Status
 
 
 def import_dummy_data():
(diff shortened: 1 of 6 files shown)


traitcuration/traits/templates/traits/datasources.html
```diff
@@ -2,8 +2,14 @@
...
-  <a href="{% url 'fetch_data' %}"
-    ><button class="uk-button uk-button-primary">FETCH DATA</button>
+  <a href="{% url 'clinvar_data' %}"
+    ><button class="uk-button uk-button-primary">CLINVAR DATA</button>
+  </a>
+  <a href="{% url 'zooma_suggestions' %}"
+    ><button class="uk-button uk-button-primary">ZOOMA SUGGESTIONS</button>
+  </a>
+  <a href="{% url 'dummy_data' %}"
+    ><button class="uk-button uk-button-primary">IMPORT DUMMY DATA</button>
```
tskir, line 12: I suggest moving this button so that it comes first in the page, because otherwise my reflex is to click them in order, in which case I'll import ClinVar and ZOOMA, and after that drop all that with dummy data import

tskir: Actually, if it's not too hard to implement, I'd like to request for a "IMPORT ALL button", which would trigger, in order:
* Dummy data import
* ClinVar import
* ZOOMA import

traitcuration/traits/datasources/dummy.py
```diff
@@ -0,0 +1,148 @@
...
+    trait_obsolete.save()
+    trait_deleted = Trait(name='Spastic paraplegia', status='deleted', number_of_source_records=7)
+    trait_deleted.save()
+    # USERS
+    user1 = User(username='user1', email='user1@user.com')
+    user1.save()
+    user2 = User(username='user2', email='user2@user.com')
+    user2.save()
+    user3 = User(username='user3', email='user3@user.com')
+    user3.save()
```
tskir, line 78: Even though this file is only for development, I would still suggest to improve readability by getting rid of repeated `save()` calls. Like this for users, and it can also be done in a similar way for other models:
```suggestion
    # USERS
    user1 = User(username='user1', email='user1@user.com')
    user2 = User(username='user2', email='user2@user.com')
    user3 = User(username='user3', email='user3@user.com')
    for user in (user1, user2, user3):
        user.save()
```

tskir: (While in this simplest case it will only save one line, for other models it will save dozens and improve readability)

traitcuration/traits/datasources/dummy.py
```diff
@@ -0,0 +1,148 @@
...
+    term_nc2.save()
+    term_ac1 = OntologyTerm(label='Diastrophic dysplasia',
+                            description='Description for Diastrophic dysplasia',
+                            cross_refs="", status='awaiting_creation')
+    term_ac1.save()
+    term_obsolete = OntologyTerm(label='Diastrophic dysplasia', curie='Orphanet:628',
+                                 iri='http://www.orpha.net/ORDO/Orphanet_628', status='obsolete')
+    term_obsolete.save()
+    term_deleted = OntologyTerm(label='Spastic paraplegia', curie='HP:0001258',
+                                iri=' http://purl.obolibrary.org/obo/HP_0001258', status='deleted')
```
tskir, line 47: Is the space at the beginning of IRI intentional to make it invalid? I think this might be problematic. For example, `requests` library for Python apparently strips leading/trailing spaces from addresses. I suspect _some_ other services _might_ do the same, hence the behaviour might not be consistent.

I suggest another way of making the term appear "deleted": simply using an impossibly large HP ID but an otherwise correct IRI. This way, when we implement OLS updates further down the road, this dummy trait will never lose its deleted status:

```suggestion
    # CURIE and IRI are intentionally incorrect so that the term stays deleted after OLS updates
    term_deleted = OntologyTerm(label='Spastic paraplegia', curie='HP:999999999',
                                iri='http://purl.obolibrary.org/obo/HP_999999999', status='deleted')
```

traitcuration/traits/datasources/dummy.py
tskir, line 78: Even though this file is only for development, I would still suggest to improve readability by getting rid of repeated `save()` calls. Granted, in _this_ case it only saves one line, but if this is done in a similar manner for other models (which I suggest you do), it will improve readability a lot:

```suggestion
    # USERS
    user1 = User(username='user1', email='user1@user.com')
    user2 = User(username='user2', email='user2@user.com')
    user3 = User(username='user3', email='user3@user.com')
    for user in (user1, user2, user3):
        user.save()
```

traitcuration/traits/templates/traits/datasources.html
tskir, line 12: I suggest moving this button so that it comes first in the page, because otherwise my reflex is to click them in order, in which case I'll import ClinVar and ZOOMA, and then proceed to drop all that with dummy data import :)

Also, if it's not too hard to implement, I'd like to ask for a "IMPORT ALL button", which would trigger, in order: 
* Dummy data import
* ClinVar import
* ZOOMA import

traitcuration/traits/datasources/dummy.py
```diff
@@ -0,0 +1,148 @@
...
+
+def import_dummy_data():
+    Review.objects.all().delete()
+    MappingSuggestion.objects.all().delete()
+    Mapping.objects.all().delete()
+    Trait.objects.all().delete()
+    OntologyTerm.objects.all().delete()
+    User.objects.all().delete()
+    # ONTOLOGY TERMS
+    term_current1 = OntologyTerm(label='Diabetes mellitus', curie='EFO:0000400',
```
tskir, line 15: I suggest renaming the dummy traits so that it is obvious they do not really come from ClinVar and have been pre-populated with data. This can be done by e.g. naming them '/ Diabetes mellitus /' or in some other way. This will not prevent them from being updated by ZOOMA, but will clearly mark and distinguish them.

traitcuration/traits/datasources/dummy.py
```diff
@@ -0,0 +1,148 @@
...
+    term_nc2 = OntologyTerm(label='Familial cancer of breast, 2',
+                            description='Description for familial cancer of breast, 2',
+                            cross_refs="Orphanet:0000405", status='awaiting_creation')
+    term_nc2.save()
+    term_ac1 = OntologyTerm(label='Diastrophic dysplasia',
+                            description='Description for Diastrophic dysplasia',
+                            cross_refs="", status='awaiting_creation')
+    term_ac1.save()
+    term_obsolete = OntologyTerm(label='Diastrophic dysplasia', curie='Orphanet:628',
+                                 iri='http://www.orpha.net/ORDO/Orphanet_628', status='obsolete')
```
tskir, line 44: Suggest replacing this with an actual obsolete term, like http://www.ebi.ac.uk/efo/EFO_0003093

joj0s:
Changes added:
- Replace dummy obsolete trait with actually obsolete trait
- Assign broken iri to dummy deleted term 
- Move saving inside for loops
- Add 'import all data' button
- Create dummy mapping suggestions

traitcuration/traits/datasources/zooma.py
```diff
@@ -95,7 +94,10 @@ def create_mapping_suggestion(trait, term):
     """
     Creates a mapping suggestion in the app's database, if it doesn't exist already.
     """
-    zooma = User.objects.filter(username="Zooma").first()
+    zooma = User.objects.filter(username="ZOOMA").first()
+    if zooma is None:
+        zooma = User(username="ZOOMA", email="zooma@zooma.com")
```
tskir, line 99: Since zooma.com is a real domain (and not relevant to our ZOOMA as well), it's better to substitute the development mailing list of EVA

```suggestion
        zooma = User(username="ZOOMA", email="eva-dev@ebi.ac.uk")
```

PR #60, opened 2020-07-07
Background tasks using Celery

### LINKED ISSUE #59: ZOOMA import times out intermittently
I noticed that in deployment #51, but it will probably be true for any recent deployment.

Currently, ZOOMA query to update suggestions for all mappings (dummy + a selection of ClinVar which we import) takes just about 30 seconds to complete. Due to randomness in ZOOMA server response, sometimes it's less than 30s and everything's fine. However, when the time exceeds 30 seconds, the request times out:
image

```
[...]
2020-07-04T13:51:56.363771+00:00 app[web.1]: INFO:ZOOMA:Deleted mapping suggestions <QuerySet []>
2020-07-04T13:51:56.363905+00:00 app[web.1]: INFO:ZOOMA:Retrieving ZOOMA suggestions for trait: Lung adenocarcinoma
2020-07-04T13:51:56.400109+00:00 app[web.1]: [2020-07-04 13:51:56 +0000] [4] [CRITICAL] WORKER TIMEOUT (pid:14)
2020-07-04T13:51:56.401534+00:00 app[web.1]: [2020-07-04 13:51:56 +0000] [14] [INFO] Worker exiting (pid: 14)
```

To solve that, we will need to implement all long queries (ClinVar, ZOOMA etc.) to run in background, rather than waiting for them to complete inside a single request. Hopefully Heroku supports running something on the server in the background—otherwise we'll have to prioritise setting up another means of deployment.

joj0s:
This PR adds the ability to run background tasks using Celery. It is marked as WIP until tests on Heroku are done. Closes #59


traitcuration/traits/datasources/clinvar.py
```diff
@@ -16,6 +16,19 @@
...
+def run_clinvar():
+    """
+    This function initiates the process of downloading, parsing and storing ClinVar data
+    """
+    try:
+        download_clinvar_data()
+        traits_dict = parse_trait_names_and_source_records()
+        store_data(traits_dict)
+    except Exception as e:
+        print(f"Error: {e}")
```
tskir, line 28: Does this print the entire stack trace, or just the exception type & message? Ideally we want the stack trace to be able to understand what happened.

traitcuration/traits/tasks.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import absolute_import
```
tskir, line 1: Wait... Aren't we using Python 3 for all of this? Why is this import from `__future__` necessary?

joj0s:
> Looks good (left two minor comments), but I understand this won't work on Heroku due to what you described in #59, right?

Yes, this is why I marked this PR as work in progress, I am still testing things with Heroku, and I still have some local uncommitted code. I will use the "Request Review" feature and umnark this as WIP  to notify you when it is ready for submission

traitcuration/traits/tasks.py
joj0s: Yeah it was a recommended solution for a problem I was having with Celery finding some dependencies, it will be removed in the final iteration of this PR

traitcuration/traits/datasources/clinvar.py
joj0s: Agreed, dded stack trace printing

joj0s:
@tskir The PR is now ready for submission. It works on Heroku due to a Redis instance of mine that I attached so feel free to test it out and check out the logs of the instance. I also added a progress bar for ZOOMA imports, mainly used for development, as I think it would be better to overhaul the 'datasources' page once we have the main external data sources functionality ready.

traitcuration/traits/tasks.py
tskir: Looks like imports from the `__future__` are still there? I can see them in the `traitcuration/__init__.py` and `traitcuration/celery.py` modules

tskir:
> It works on Heroku due to a Redis instance of mine that I attached so feel free to test it out and check out the logs of the instance.

Okay, that's great! I'm just a bit confused now as to how this relates to the issues we discussed in #59. I thought it was a necessity to attach billing information to the specific project the app is running in? Did you manage to overcome that in some way?

joj0s:
>  I'm just a bit confused now as to how this relates to the issues we discussed in #59. I thought it was a necessity to attach billing information to the specific project the app is running in? Did you manage to overcome that in some way?

So, I have an empty personal project of mine and a Redis instance attached to it. If you look at settings.py, I have hardcoded the URL of the Redis instance running there, so the app connects to that.

tskir:
Ah, yes, I now see that in the settings. For development purposes this is (marginally) acceptable, but in general is dangerous. Let's just hope it won't get abused by some bot crawling the web for Redis URLs (for some reason). I will expedite the process of procuring a proper environment to avoid having to use that for extended periods of time.

joj0s:
You're right about the URL, I should probably set it back as an env variable and then manually insert it in the Heroku environment. I'll do that and then merge this

traitcuration/traits/tasks.py
joj0s: Oh my bad, I forgot I included them in other files besides `tasks.py`. I removed them and pushed the changes.

tskir:
Okay that would be great!

PR #62, opened 2020-07-18
Implement authentication

### LINKED ISSUE #54: Implement user authentication
Based on decisions made in #53.

joj0s:
This PR adds authentication of users via the Google OAuth API. Closes #54.


traitcuration/settings.py
```diff
@@ -44,11 +44,18 @@
...
     'django_celery_results',
-    'celery_progress'
+    'celery_progress',
+    'allauth',
+    'allauth.account',
+    'allauth.socialaccount',
+    'allauth.socialaccount.providers.google',
 ]
 
+SITE_ID = 2
```
tskir, line 57: What does this do, and why is this “2”?

traitcuration/settings.py
```diff
@@ -89,6 +96,34 @@
...
+}
+
+AUTHENTICATION_BACKENDS = [
+    # Needed to login by username in Django admin, regardless of `allauth`
+    'django.contrib.auth.backends.ModelBackend',
+    # `allauth` specific authentication methods, such as login by e-mail
+    'allauth.account.auth_backends.AuthenticationBackend',
+]
+
+ACCOUNT_EMAIL_REQUIRED = True
```
tskir, line 121: Perhaps it's worth mentioning in a comment that those configuration options are of the “allauth” package

traitcuration/traits/datasources/dummy.py
```diff
@@ -54,10 +54,10 @@ def import_dummy_data():
     for trait in (trait1, trait2, trait3, trait4, trait5, trait6, trait7, trait8, trait9, trait10):
         trait.save()
     # USERS
-    user1 = User(username='/ user1 /', email='user1@user.com')
-    user2 = User(username='/ user2 /', email='user2@user.com')
-    user3 = User(username='/ user3 /', email='user3@user.com')
-    user4 = User(username="ZOOMA", email="zooma@zooma.com")
+    user1 = User(email='user1@user.com', first_name="John", last_name="Doe")
```
tskir, line 57: Just to confirm, does Google OAuth return first and last names separately, right? (I can't seem to be able to find the part in the code where the OAuth reply is being actually converted into our user model)

traitcuration/traits/datasources/zooma.py
```diff
@@ -98,9 +98,9 @@ def create_mapping_suggestion(trait, term):
     """
     Creates a mapping suggestion in the app's database, if it doesn't exist already.
     """
-    zooma = User.objects.filter(username="ZOOMA").first()
+    zooma = User.objects.filter(first_name="ZOOMA").first()
```
tskir, line 101: Filtering by the name seems a bit wrong. I'm not saying we are likely to encounter someone with the first name “ZOOMA”, but maybe it would be better to filter by email

traitcuration/traits/managers.py
```diff
@@ -0,0 +1,35 @@
+from django.contrib.auth.base_user import BaseUserManager
+from django.utils.translation import ugettext_lazy as _
+
+
+class CustomUserManager(BaseUserManager):
+    """
+    Custom user model manager where email is the unique identifiers
+    for authentication instead of usernames.
```
tskir, line 8: This can fit into one line <120 characters. Looks like pylint strikes again :)

```suggestion
    Custom user model manager where email is the unique identifier for authentication instead of usernames.
```

traitcuration/traits/models.py
```diff
@@ -37,13 +41,20 @@ def __str__(self):
         return self.name
 
 
-class User(models.Model):
-    username = models.CharField(max_length=100, unique=True)
-    email = models.EmailField(unique=True)
-    role = models.CharField(max_length=20, blank=True, null=True)
```
tskir, line 43: Looks like we lost the "role" field in the refactoring. Or is it included in some form in the AbstractUser and inherited from there?

traitcuration/traits/templates/traits/base.html
```diff
@@ -35,11 +37,23 @@
...
+                        <span uk-icon="google" class="uk-margin-small-right"> </span> LOG IN 
+                    </a> 
+                </button>
+            {% else %}
+                <div class="auth-links">
+                    <span uk-icon="icon: user; ratio: 1.3" class="auth-links__icon"></span>
+                    <span>{% user_display user %}</span>
+                    <button class="uk-button button button-outlined--transparent">
+                        <a href="/accounts/logout/?next={{request.path}}">
+                            <span uk-icon="sign-out" class="uk-margin-small-right"> </span> SIGN OUT
```
tskir, line 52: When logged in, the email is too close to the "Sign out" button:

image

tskir:
Also, I tried refreshing the ZOOMA mappings, and I'm getting this error, could you look into it?

image

traitcuration/traits/datasources/dummy.py
joj0s: It does automatically, yes.

traitcuration/traits/datasources/zooma.py
joj0s: I actually added that in PR #63 so should I leave it as it is and wait for both PRs to merge into master?

traitcuration/traits/templates/traits/base.html
joj0s: Same problem with PR #63 , looks like the `gap` CSS property isn't fully supported. May I ask what browser are you using?

traitcuration/settings.py
joj0s: Agreed, I will add that.

traitcuration/settings.py
joj0s: So in order for the auth package to work, I have to register the domain in the app's database, you can see it in the `Site` model in the Django Admin. It already has an example site there which I left as is during development and added another entry with the actual domain, hence the id of 2. I forgot I still had that as 2 so I will change it back to 1 and modify the database in the deployment accordingly

joj0s:
> Also, I tried refreshing the ZOOMA mappings, and I'm getting this error, could you look into it?

I'm not sure what happened, I just ran the ZOOMA suggestion retrieval by itself and it ran fine. Did you make any other changes to the database since that error occurred?

traitcuration/traits/models.py
joj0s: You are right, I wll add that.

traitcuration/settings.py
tskir: If it's not too difficult, let's change it to 1 and get rid of the example site entities. Also, could you please add your explanation (or maybe just the gist of it) as a comment for this `SITE_ID` assignment?

traitcuration/traits/datasources/zooma.py
tskir: Yes, if it's already planned to be fixed in a subsequent PR, then that's fine, no need to backport the fix here

traitcuration/traits/templates/traits/base.html
tskir: Looks like you fixed it, since now it looks good:
image

I'm always using the latest(ish) release Google Chrome on Ubuntu. My current version is: 83.0.4103.97 (Official Build) (64-bit)

PR #63, opened 2020-07-20
Suggestion of existing or new terms

### LINKED ISSUE #26: Implement the “Suggest existing term” section
image

This should include some form of validation (the pasted URI should point to a valid ontology).

### LINKED ISSUE #27: Implement the “Create new term” section
image

As we discussed, the current approach is to use a pop-up window. This should include trait name, description, and database cross-references.

joj0s:
This PR is built on top of PR #62 and adds the ability to suggest new or existing terms. Closes #26, closes #27.

### DIFF
--- a/traitcuration/static/js/mapping.js
+++ b/traitcuration/static/js/mapping.js
@@ -3,7 +3,7 @@ This module contains all the funcionality for mapping traits to ontology terms, 
 */
 let selectedRowIndex = -1;
 let selectedTableId = "";
-let currentTraitId = -1;
+let currentTraitId = window.location.href.split("/").slice(-2)[0];
 let selectedTermId = -1;
 
 /* 
@@ -27,14 +27,65 @@ function selectRow(row, tableId, traitId, termId) {
   selectedRow.classList.add("suggestion-table__row--selected");
 }
 
-// This function makes an ajax request to create a current mapping with the selected term 
+
+// This function makes an ajax request to create a current mapping with the selected term
 function mapButtonClicked() {
   if (selectedRowIndex < 1) return;
   axios
-    .post(`/traits/${currentTraitId}/mapping`, {'term': parseInt(selectedTermId)})
+    .post(`/traits/${currentTraitId}/mapping`, {
+      term: parseInt(selectedTermId),
+    })
     .then((response) => {
       // handle success
       console.log(response);
-      location.reload(); 
+      location.reload();
     });
+}
+
+
+/* This function accepts a term IRI pasted by a user and if it doesn't already exist, creates a new mapping suggestion
+and mapping for that term */
+async function existingTermButtonClicked() {
+  // Check if the IRI field is empty
+  termIRI = document.getElementById("existingTermIRI").value;
+  if (termIRI === "") return;
+  // Check if a mapping suggestion with the given IRI already exists in the suggestions
+  for (const row of document.querySelectorAll(".suggestion-table__link")) {
+    if (row.getAttribute('href') === termIRI) {
+      UIkit.notification({
+        message: "A suggestion of this term already exists",
+        status: "warning",
+        pos: "top-center",
+        timeout: 3000,
+      })
+      return;
+    }
+  }
+  /*
+    Extract the ontology id from the term iri, to be used for OLS queries by reading the last part of an iri and
+    reading the ontology id using the term prefix
+    E.g. extract 'mondo' from http://purl.obolibrary.org/obo/MONDO_0019482
+  */
+  termOntologyId = termIRI.split("/").slice(-1)[0].split("_")[0].toLowerCase();
+  try {
+    await axios.get(
+      `https://www.ebi.ac.uk/ols/api/ontologies/${termOntologyId}/terms?iri=${termIRI}`
+    );
+    axios
+      .post(`/traits/${currentTraitId}/mapping/add`, { term: termIRI })
+      .then((response) => {
+        // handle success
+        console.log(response);
+        location.reload();
+      });
+  } catch (err) {
+    UIkit.notification({
+      message: "Error verifying term! Check if given IRI is valid",
+      status: "danger",
+      pos: "top-center",
+      timeout: 3000,
+    });
+  }
+  // .then()
+  // .catch( err => console.log(err));
 }
--- a/traitcuration/static/js/mapping.js
+++ b/traitcuration/static/js/mapping.js
@@ -3,9 +3,14 @@ This module contains all the funcionality for mapping traits to ontology terms, 
 */
 let selectedRowIndex = -1;
 let selectedTableId = "";
-let currentTraitId = window.location.href.split("/").slice(-2)[0];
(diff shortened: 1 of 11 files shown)


tskir:
I've changed the base branch to "auth" (#62) for now, to remove redundant diffs and make the review easier. Before merging, I'll change this back to "master"

traitcuration/static/js/mapping.js
```diff
@@ -3,13 +3,21 @@ This module contains all the funcionality for mapping traits to ontology terms,
 */
 let selectedRowIndex = -1;
 let selectedTableId = "";
-let currentTraitId = -1;
+let currentTraitId = window.location.href.split("/").slice(-2)[0];
```
tskir, line 6: Here it would be nice to add a comment explaining what we're doing and an example (full string and the result of extraction).

But also, I wonder if there are more elegant ways to pass entity attributes from Django to the JavaScript code? Perhaps including the trait ID as an attribute into the page HTML would work?

tskir: I also think I've seen examples on some websites where they embed a whole JSON with machine-readable information into the page, for the JavaScript to utilise

traitcuration/static/js/mapping.js
```diff
@@ -3,13 +3,21 @@ This module contains all the funcionality for mapping traits to ontology terms,
...
-suggestion, and sets them as selected
-*/
+
+// Check if the "Newly suggested terms" table is empty, and hide it if it is.
+const NewTermSuggestionTable = document.querySelector('#newSuggestionTable')
+console.log(NewTermSuggestionTable.rows.length)
+if (NewTermSuggestionTable.rows.length === 1) {
+  NewTermSuggestionTable.classList.add('hidden')
+  document.querySelector('#newSuggestionTable-title').classList.add('hidden')
+}
```
tskir, line 16: Wouldn't it possible to do it directly with Django while generating a page?

tskir:
With this deployment, clicking on "Import all data" results in a timeout:
image

This doesn't seem to happen in the deployment in #62, although it exhibits another problem with external datasources. I'm not sure whether the problems are related.

traitcuration/static/js/mapping.js
```diff
@@ -27,14 +35,74 @@ function selectRow(row, tableId, traitId, termId) {
...
     });
 }
+
+
+/* This function accepts a term IRI pasted by a user and if it doesn't already exist, creates a new mapping suggestion
+and mapping for that term */
+async function existingTermButtonClicked() {
+  // Check if the IRI field is empty
+  termIRI = document.getElementById("existingTermIRI").value;
+  if (termIRI === "") return;
```
tskir, line 58: Seems to me that this should also generate a warning popup, similarly to how it's done just below

traitcuration/traits/templates/traits/trait_detail.html
```diff
@@ -76,47 +75,132 @@
...
+            </div>
+            <div class="field__container">
+              <label for="{{ new_term_form.cross_refs.id_for_label }}">Cross references:</label>
+              {{ new_term_form.cross_refs }}
+            </div>
+          </div>
+          <input type="submit" value="SUBMIT AND MAP" class="button button--primary uk-button new_term__button">
+        </form>
+      </div>
+    </section>
```
tskir, line 201: Formatting of this section currently looks a bit cramped:
image

I would suggest adding more space between all elements. Especially the two section for the "existing" and "new" terms should be well separated.

traitcuration/traits/views.py
```diff
@@ -50,6 +54,34 @@ def update_mapping(request, pk):
     return HttpResponse(json.dumps(model_to_dict(mapping)), content_type="application/json")
 
 
+@login_required
```
tskir, line 57: When I try to do those actions without logging in, I get redirected to a strangely (un)formatted page:

image

tskir: It even shows me as being successfully signed in, and when I click on Sign in above, it doesn't ask me for my auth information again. So it just looks like some automatic redirect which is supposed to work doesn't.

As one solution to this problem, we could simply disable all of the buttons when logged out.

traitcuration/traits/templates/traits/trait_detail.html
joj0s: I am using the `gap` property to space those out, but it turns out that browser support for that is very limited so I will change it back to having `margin`

traitcuration/static/js/mapping.js
joj0s: I could pass a suggestion count in the page context during rendering, but I didn't want to make the view code too large. Should I do it this way?

traitcuration/static/js/mapping.js
joj0s: I will add that, yes.

traitcuration/static/js/mapping.js
joj0s: Well what I could do, is pass some variables to the page via the Django context, and then declare some global JavaScript variables withing the html where they are subsequently passed to. That being said, global JS variables are considered a pretty bad practice so I try to avoid them. Or I could declare some setter methods for passing data via a Django template.

traitcuration/traits/views.py
joj0s: > This doesn't seem to happen in the deployment in #62, although it exhibits another problem with external datasources. I'm not sure whether the problems are related.

That is because the Celery background worker isn't running in this deployment. You can verify it by checking this panel in the app's main Heroku page 
image

I have ran out of free Redis instances to deploy so one thing I could do is have the two open PRs share a Redis database, and manually choose which worker we want to run, based on what app you want to review

joj0s:
> This doesn't seem to happen in the deployment in #62, although it exhibits another problem with external datasources. I'm not sure whether the problems are related.

That is because the Celery background worker isn't running in this deployment. You can verify it by checking this panel in the app's main Heroku page 
image

I have ran out of free Redis instances to deploy so one thing I could do is have the two open PRs share a Redis database, and manually choose which worker we want to run, based on what app you want to review

traitcuration/traits/views.py
joj0s: I will remove the decorator and perform the check manually

traitcuration/static/js/mapping.js
tskir: I think the last approach is the best one, directly interpolating the values into the JavaScript code. Let's go with that

traitcuration/static/js/mapping.js
tskir: Oh, I just realised something. My original idea was to not include this table into a page _at all_ when there are no suggestions to show. But then obviously, new suggestions can be added by a user, so we still need this table to be present there even when it's empty.

So this was my mistake, and no action is required here.

traitcuration/traits/views.py
tskir: I see that now there are no editing buttons when a user is logged out, so the problem is solved

PR #70, opened 2020-07-30
Trait history section and date object formatting

### LINKED ISSUE #28: Implement the history section
image

This issue does not include comments functionality.

### LINKED ISSUE #29: Implement the comments section
image

I guess eventually we would want them to support Markdown. If it's fairly trivial (e.g. if there are existing modules which are easily to set up), this could be implemented as part of this issue. If not, please let me know and we'll add a separate issue for Markdown support.

### LINKED ISSUE #57: Configure timezones
We need to configure the server's timezone and also translate datetime objects to the end user's local timezone

joj0s:
This PR adds a history section that contains information about comments, reviews and mappings of a trait. Also adds date formatting and outputting in the browser's timezone. Closes #28, closes #29, closes #57.

### DIFF
--- a/traitcuration/static/js/time-helper.js
+++ b/traitcuration/static/js/time-helper.js
@@ -1,6 +1,7 @@
 // This module contains utilities for date formatting and output.
 
-const timeFormat = "MMMM Do YYYY, H:MM ";
+
+const timeFormat = "MMMM Do YYYY, H:mm"; // e.g. July 29th 2020, 14:25
 
 // Uses moment.js library to format a ISO formatted date string
 function formatDate(isoformat_timestamp) {
--- a/traitcuration/static/js/time-helper.js
+++ b/traitcuration/static/js/time-helper.js
@@ -1,6 +1,6 @@
 // This module contains utilities for date formatting and output.
 
-const timeFormat = "MMMM Do YYYY, h:mm a";
+const timeFormat = "MMMM Do YYYY, H:MM ";
 
 // Uses moment.js library to format a ISO formatted date string
 function formatDate(isoformat_timestamp) {
--- a/traitcuration/static/js/time-helper.js
+++ b/traitcuration/static/js/time-helper.js
@@ -1,0 +1,19 @@
+// This module contains utilities for date formatting and output.
+
+const timeFormat = "MMMM Do YYYY, h:mm a";
+
+// Uses moment.js library to format a ISO formatted date string
+function formatDate(isoformat_timestamp) {
+  const momentObj = moment(isoformat_timestamp);
+  return momentObj.format(timeFormat);
+}
+
+
+// Queries elements containing dates and appends the formatted date strings to their HTML
+function outputDates() {
+  const elements = document.querySelectorAll(".date");
+
+  for (const element of elements) {
+    element.innerHTML = formatDate(element.getAttribute("date"));
+  }
+}
--- a/traitcuration/traits/templates/traits/trait_detail.html
+++ b/traitcuration/traits/templates/traits/trait_detail.html
@@ -259,28 +259,38 @@
 
       {% for event in history_events|dictsortreversed:"date" %}
       {% if event.type == 'comment' %}
-      <div class="event event--comment">
-        <span class="event__title">{{event.content.author.get_full_name}} commented</span>
-        <span date="{{event.date.isoformat}}" class="date event__date"></span>
-        <p>{{event.content.body}}</p>
+      <div class="event"">
+        <p> 
+          <span class="bold" > {{event.content.author.get_full_name}}</span> commented on 
+          <span date="{{event.date.isoformat}}" class="date bold"></span>
+        </p>
+        <p>
+          {{event.content.body}}
+        </p>
       </div>
 
       {% elif event.type == 'review' %}
-      <div class="event">
-        <p class="event__title"> 
-          {{event.content.reviewer.get_full_name}} reviewed mapping of trait "{{trait.name}}"
-          to term "{{event.content.mapping_id.term_id.curie}} - {{event.content.mapping_id.term_id.label}}"
+      <div class="event"">
+        <p> 
+          <span class="bold" > {{event.content.reviewer.get_full_name}}</span> confirmed the mapping on 
+          <span date="{{event.date.isoformat}}" class="date bold"></span>
         </p>
-        <span date="{{event.date.isoformat}}" class="date event__date"></span>
+        <p>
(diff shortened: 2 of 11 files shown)


traitcuration/static/js/time-helper.js
```diff
@@ -0,0 +1,19 @@
+// This module contains utilities for date formatting and output.
+
+const timeFormat = "MMMM Do YYYY, H:MM ";
```
tskir, line 3: I can see what this looks like in the final deployment, but it would be helpful to also include a comment with an example of the final formatted timestamp here in the code

traitcuration/traits/templates/traits/trait_detail.html
```diff
@@ -236,11 +236,63 @@ <h4>New Term Information</h4>
       </div>
     </section>
     {% endif %}
+
+    <section class="section">
+      <h4>TRAIT HISTORY</h4>
```
tskir, line 241: Two comments about the trait history header section:
image

1. For individual comments the left border looks good as an UI element. However, I think that for the “Comment as” section we should get rid of it, and align it on the same level with the "TRAIT HISTORY" title, in order to distinguish it from the actual comments and historical actions below.
2. As you can see, for some reason the input field is rendered in such a way that it has a tiny scrollbar. The styles probably need to be tweaked a bit to get rid of it. This might be something specific to Chrome.

traitcuration/traits/templates/traits/trait_detail.html
```diff
@@ -236,11 +236,63 @@ <h4>New Term Information</h4>
...
+
+      {% if not user.is_anonymous %}
+      <div class="post-comment__container">
+        <div class="post-comment__header">
+          <span uk-icon="user"></span>
+          <span>Comment as {{user.get_full_name}}</span>
+          <button class="uk-button button button--primary post-comment__submit" onclick="commentButtonClicked()">Submit</button>
+        </div>
```
tskir, line 249: I feel that the representation of text comments could be made more compact. Currently we have:

image

**Kirill Tsukanov commented**  
August 1st 2020, 23:08  
(blank line)  
(Comment text)  
(blank line)

This can be shortened to:

**Kirill Tsukanov** commented on **August 1st 2020, 23:08**  
(Comment text)

traitcuration/traits/templates/traits/trait_detail.html
```diff
@@ -236,11 +236,63 @@ <h4>New Term Information</h4>
...
+
+        <div class="post-comment__body">
+          <textarea 
+          id="commentBody" 
+          class="uk-textarea post-comment__textarea" 
+          placeholder="Leave a commment..."></textarea>
+        </div>
+      </div>
+      {% endif%}
```
tskir, line 258: image

For the mapping change entry, I feel this could be changed to (note the em dash and making the CURIE part clickable):

**Kirill Tsukanov** proposed a mapping on **August 1st 2020, 23:08**  
**[Orphanet:101016](http://www.orpha.net/ORDO/Orphanet_101016) — Romano-Ward syndrome**

traitcuration/traits/templates/traits/trait_detail.html
```diff
@@ -236,11 +236,63 @@ <h4>New Term Information</h4>
...
+
+      {% for event in history_events|dictsortreversed:"date" %}
+      {% if event.type == 'comment' %}
+      <div class="event event--comment">
+        <span class="event__title">{{event.content.author.get_full_name}} commented</span>
+        <span date="{{event.date.isoformat}}" class="date event__date"></span>
+        <p>{{event.content.body}}</p>
+      </div>
```
tskir, line 266: When proposing new terms, they currently display like this:
image

It probably shouldn't display "None" explicitly

traitcuration/traits/templates/traits/trait_detail.html
```diff
@@ -236,11 +236,63 @@ <h4>New Term Information</h4>
...
+
+      {% elif event.type == 'review' %}
+      <div class="event">
+        <p class="event__title"> 
+          {{event.content.reviewer.get_full_name}} reviewed mapping of trait "{{trait.name}}"
+          to term "{{event.content.mapping_id.term_id.curie}} - {{event.content.mapping_id.term_id.label}}"
+        </p>
+        <span date="{{event.date.isoformat}}" class="date event__date"></span>
+      </div>
```
tskir, line 275: For the review entries:
image

This can be changed to:

**Kirill Tsukanov** confirmed the mapping on **August 1st 2020, 23:08**  
**[Orphanet:586](http://www.orpha.net/ORDO/Orphanet_586) — Cystic fibrosis**

traitcuration/traits/templates/traits/trait_detail.html
```diff
@@ -236,11 +236,63 @@ <h4>New Term Information</h4>
...
+          {{event.content.curator.get_full_name}} mapped trait 
+          "{{trait.name}}" to term "{{event.content.term_id.curie}} - {{event.content.term_id.label}}"
+        </p>
+        <span date="{{event.date.isoformat}}" class="date event__date"></span>
+      </div>
+
+      {% endif %}
+
+      {% endfor %}
+    </section>
```
tskir, line 289: Finally, there appears to be this strange bug where the timestamp of the comment does not actually reflect the time when I leave it, and instead reflects the time the app was first launched (?)

For example, I started reviewing this PR at 23:08 and opened the deployment. Since then, all comments which I leave have the timestamp of 23:08, for example:

image

traitcuration/traits/templates/traits/trait_detail.html
joj0s: Should I just display the label then?

traitcuration/traits/templates/traits/trait_detail.html
joj0s: That is indeed a weird Chrome specific issue and it looks like quite a pain because if I just disable the scrollbar, then the user wouldn't be able to scroll up an down if the comment size exceeds the text area height. So I think the best approach here would be to hide the scrollbar when the text area is collapsed, and show it when it is expanded.

tskir: Yes, this approach works great, I just tested it. I'll resolve this conversation then, great job

traitcuration/traits/templates/traits/trait_detail.html
tskir: Yes, something like:

> **Kirill Tsukanov** proposed a mapping to a new term on **August 4th 2020, 8:59**
> 
> **New term label**

Without the link

tskir: Currently it contains a broken link to "None", like that:
image

Issue #58, opened 2020-07-04
Production launch checklist


tskir:
This meta-issue tracks important things which we shouldn't forget to enable or implement before rolling the app into production once it is ready.

- [ ] Ensure CSRF protection is enabled back

joj0s:
There is also this list that will help
https://docs.djangoproject.com/en/3.0/howto/deployment/checklist/

PR #77, opened 2020-08-03
Query OLS to update term status

### LINKED ISSUE #31: Implement querying OLS to update term status
Example of an OLS query through JSON API: https://www.ebi.ac.uk/ols/api/ontologies/efo/terms?iri=http://www.ebi.ac.uk/efo/EFO_0000612

This ticket only inlcudes developing the capability to do the query and update the data. Running it automatically or manually will be implemented in other tickets.

joj0s:
This PR adds the ability to query OLS for status updates, via asynchronous background tasks. Closes #31.


joj0s:
@tskir testing this functionality is tricky, because it turns out that `django-computedfields` doesn't play well with the Django admin site. Changing a status value directly from the admin site causes an error in the term. So if you want I could write a script that changes some status values of terms via the Django shell, so that you can test if they are updated correctly when the OLS queries are made.

traitcuration/traits/datasources/ols.py
```diff
@@ -7,8 +7,11 @@
 
 from retry import retry
 
+from ..models import Status, OntologyTerm
+
+
 logging.basicConfig()
-logger = logging.getLogger('ZOOMA')
+logger = logging.getLogger('OLS')
```
tskir, line 14: FYI it is a common practice to just always use `logger = logging.getLogger(__name__)` in order to avoid explicitly specifying module name each time. This is not mandatory to adopt, but can be useful

traitcuration/traits/tasks.py
```diff
@@ -26,3 +26,10 @@ def get_clinvar_data_and_suggestions():
     clinvar.run_clinvar()
     zooma.run_zooma_for_all_traits()
     return 'Successful Import'
+
+
+@shared_task
+def get_ols_status():
+    ols.check_term_status()
+    ols.check_awaiting_import_terms()
+    return 'Successful status update'
```
tskir, line 35: I would suggest having only one call to some main function in the `ols` module here, which would do all necessary processing steps. Otherwise, if we have multiple top-level functions in that module and multiple calls here, it creates a possibility for code sync problems

traitcuration/traits/datasources/ols.py
```diff
@@ -45,3 +48,51 @@ def get_ontology_id(term_iri):
...
+        status__in=[Status.AWAITING_IMPORT, Status.NEEDS_IMPORT, Status.CURRENT])
+    for term in terms:
+        logger.info(f"Checking term {term.iri}")
+        term_ontology_id = get_ontology_id(term.iri)
+        term_info = make_ols_query(term.iri, term_ontology_id)
+        if term_info is not None:
+            term.status = get_term_status(term_info['is_obsolete'], term_ontology_id)
+        else:
+            term.status = Status.DELETED
+        term.save()
```
tskir, line 98: It's a question of personal preference, but I feel it would be better to have just one function here which would, in one cycle, go through all traits and make the necessary queries/status updates. That would make the logic more readable.

Also, I see that this currently the “current” terms are only queried in their native ontology, and only “awaiting import” are queried in EFO. But actually, we want to query _all_ terms both in their native ontology and in EFO (of course, if EFO is the native ontology, then we only want to make one query), and then make the appropriate decisions.

traitcuration/traits/views.py
```diff
@@ -138,3 +138,8 @@ def zooma_suggestions(request):
 def dummy_data(request):
     dummy.import_dummy_data()
     return redirect('datasources')
+
+
+def ols_queries(request):
+    get_ols_status.delay()
+    return redirect('datasources')
```
tskir, line 145: (I'm adding this here so that we can have a thread rather than isolated comments)

> @tskir testing this functionality is tricky, because it turns out that django-computedfields doesn't play well with the Django admin site. Changing a status value directly from the admin site causes an error in the term. So if you want I could write a script that changes some status values of terms via the Django shell, so that you can test if they are updated correctly when the OLS queries are made.

I think this can be done somewhat easier by just expanding the existing “test suite” of dummy traits which we already have. We could add some dummy traits which have their status purposefully incorrect to start with, and then test the process by clicking the "OLS STATUS UPDATES". Eventually this can also evolve smoothly into an automated test suite, just as the other dummy traits.

traitcuration/traits/datasources/ols.py
joj0s: > It's a question of personal preference, but I feel it would be better to have just one function here which would, in one cycle, go through all traits and make the necessary queries/status updates. That would make the logic more readable.

I am not sure I understood that, you mean we should merge the `check_awaiting_import_terms` and the `check_term_status` functions?

tskir: Yes, exactly. Into one function which would go through all traits (regardless of their _current_ status), and then for _each_ trait do the necessary logic for queries and status updates. Something like (pseudocode):

```
def ols_update():
    for trait in all_traits:
        query efo
        query native ontology
        new_status = decide_status(current_status, efo_status, native_ontology_status)
```

tskir: In that example, the `decide_status` doesn't have to be a separate function (and probably even shouldn't be). This could all be a monolithic function which both does the queries and then immediately decides the new status of the traits based on a bunch of branching **if** statements. The purpose of this is to translate the logic of choosing the new status as directly into code as possible, so that it's easy to verify its correctness

tskir: Let me know if this is still not clear, I'll be happy to elaborate more

joj0s: Yes, I think this approach makes much more sense. Although I think it would be best then to separate the status calculation then, since there are going to be a bunch of if statements, and it is going to make the code much more readable.

tskir: > Although I think it would be best then to separate the status calculation, since there are going to be a bunch of if statements, and it is going to make the code much more readable.

Well yes, actually a good point. I agree.

jerch:
> testing this functionality is tricky, because it turns out that django-computedfields doesn't play well with the Django admin site.

@joj0s May I ask whats not working for you with `django-computedfields`? Also feel free to write issues, if you find something not working or feels like going into the wrong direction with computed fields. I still need to collect evidence where and to what extend it can be helpful at all.

joj0s:
Hi @jerch, first of all thanks for coming here and providing assistance. 

I have been investigating an issue since yesterday, where a `TextChoices` CharField which is being computed using your package (the computation works perfectly fine), suddenly breaks when it is being changed manually through Django Admin. I am still looking for the source of it, so I might have been quick to blame it on the package itself. 

Once I identify the exact problem, I can open an issue in the package's page to investigate it further, and of course provide any other feedback needed to improve it.

jerch:
> ... suddenly breaks when it is being changed manually through Django Admin

Yes, thats indeed not supported, computed fields are non editable by default (not even sure, how you got that changed) and get recalculated from the method code on `.save()`. Setting values manually to CFs has no meaning, those values would be lost.

May I ask whats your exact usecase here? For me it was never an issue to be able to set a CF value manually, furthermore it would work against the idea of computed fields, as it skips the method logic (prolly leading to desync issues).

joj0s:
@jerch Bad wording on my part, I didn't try to manually change the computed field itself, rather the field that the computed field depended on. 

I did manage to fix it though, and the mentioned problem was unrelated to the `django-computedfields` in the end, so the package is continuing to serve as well.

jerch:
@joj0s Ah ok, np.

I had a quick look at your defined CFs in the code, I see one issue with this one
https://github.com/EBIvariation/trait-curation/blob/532ef37e694fb401f32117b456881daba3fef81f/traitcuration/traits/models.py#L88-L92

The concrete field listing on the right side is empty, which basically makes this rule a NOOP (no explicit dependency is added). This might lead to desync issues, if elements of `review_set` get moved between `Mapping` instances. To avoid that, you should list the fk field pointing back on the right side as concrete source field, here `['review_set', ['mapping_id']]`.

joj0s:
@jerch Thank you very much for this suggestion, I will add that!

traitcuration/traits/views.py
joj0s: The admin issue is fixed so feel free to change whatever you like there, however I did add a dummy obsolete term (*/ obsolete_adrenocortical carcinoma / - EFO:0003093*) that is marked as 'current' and a dummy 'current' term (*/ Hereditary breast cancer / - Orphanet:227535*) that is marked as 'awaiting_import' to test the most usual cases right away.

traitcuration/traits/datasources/dummy.py
```diff
@@ -14,14 +14,14 @@ def import_dummy_data():
     Mapping.objects.all().delete()
     Trait.objects.all().delete()
     OntologyTerm.objects.all().delete()
-    User.objects.all().delete()
+    User.objects.exclude(email="admin@admin.com").delete()
```
tskir, line 17: Does this mean that we have admin@admin.com hardcoded somewhere as the email for the administrative user?

I feel like it would be better to either replace it with something sensible (e.g. eva-dev@ebi.ac.uk), or at least use the example.com domain instead of admin.com (which is a domain which actually exists)

traitcuration/traits/datasources/dummy.py
```diff
@@ -14,14 +14,14 @@ def import_dummy_data():
...
     # ONTOLOGY TERMS
     term1 = OntologyTerm(label='/ Diabetes mellitus /', curie='EFO:0000400',
                          iri='http://www.ebi.ac.uk/efo/EFO_0000400', status=Status.CURRENT)
     term2 = OntologyTerm(label='/ digestive system disease /', curie='EFO:0000405',
                          iri='http://www.ebi.ac.uk/efo/EFO_0000405', status=Status.CURRENT)
-    term3 = OntologyTerm(label='/ Insulin-resistant diabetes mellitus /', curie='HP:0000831',
-                         iri='http://purl.obolibrary.org/obo/HP_0000831', status=Status.NEEDS_IMPORT)
+    term3 = OntologyTerm(label='/ Hereditary breast cancer /', curie='Orphanet:227535',
```
tskir, line 23: It would be useful to add a comment here that the term here is purposefully incorrect to test OLS updates

traitcuration/traits/datasources/dummy.py
```diff
@@ -37,24 +37,24 @@ def import_dummy_data():
                          description='Description for Diastrophic dysplasia',
                          cross_refs="", status=Status.AWAITING_CREATION)
     term9 = OntologyTerm(label='/ obsolete_adrenocortical carcinoma /', curie='EFO:0003093',
-                         iri='http://www.ebi.ac.uk/efo/EFO_0003093', status=Status.OBSOLETE)
+                         iri='http://www.ebi.ac.uk/efo/EFO_0003093', status=Status.CURRENT)
```
tskir, line 40: Also same thing here about adding the comment for purposefully incorrect status

traitcuration/traits/datasources/dummy.py
```diff
@@ -37,24 +37,24 @@ def import_dummy_data():
...
     term10 = OntologyTerm(label='/ Spastic paraplegia /', curie='HP:999999999',
                                 iri=' http://purl.obolibrary.org/obo/HP_999999999', status=Status.DELETED)
```
tskir, line 42: Oh, by the way, another thing which can definitely change (and frequently does), is the term label. We must ensure that this is also updated on OLS queries. I believe this is currently not implemented due to an oversight from my side (I just forgot to mention this case)

tskir: If this requires fixing more than just a few lines, feel free to add a follow-up issue rather than fixing it directly in this PR

traitcuration/traits/datasources/ols.py
```diff
@@ -45,3 +49,54 @@ def get_ontology_id(term_iri):
...
+    'Obsolete' if the is_obsolete flag is true, 'Deleted' if no info was found in any ontology, 'Current' if its
+    ontology is EFO, and 'Needs Import' if info about the term was found in another ontology.
+    """
+    if ontology_id is None:
+        return Status.DELETED
+    if is_obsolete:
+        return Status.OBSOLETE
+    if ontology_id == 'efo':
+        return Status.CURRENT
+    return Status.NEEDS_IMPORT
```
tskir, line 66: The overall logic of status selection & querying OLS in a single function is now much better. However, what worries me is that we have two completely separate functions which decide the statuses: `get_term_status` (which is, by the way, defined in _ols.py_ but used only in _zooma.py_ — looks a bit odd) and `update_status`. So ultimately there are two places in the code where the term can be decided to be deleted or obsolete. Ideally, this should be done only once, and all of the different cases must use exactly the same status determination logic:
* Term is added from ZOOMA
* Term is added by user
* Term status is being updated by periodic OLS lookup

traitcuration/traits/datasources/ols.py
```diff
@@ -45,3 +49,54 @@ def get_ontology_id(term_iri):
...
+    """
+    term_ontology_response = efo_response if term.status == Status.CURRENT else parent_ontology_response
+    if term.status == Status.AWAITING_IMPORT and efo_response is not None:
+        logger.info(f"FOUND IN EFO: Term {term}")
+        return Status.CURRENT
+    else:
+        if term_ontology_response is None:
+            logger.info(f"FOUND DELETED: Term {term}")
+            return Status.DELETED
+        elif term_ontology_response['is_obsolete'] is True:
```
tskir, line 98: If I'm reading this right, it looks like we only check one ontology. However, for example, for terms which were imported from MONDO (or any other ontology), we need to check _both_ the parent ontology and the EFO. If the term is obsolete in _either,_ it should lead to it being declared obsolete in the status

joj0s: I'm kind of confused, a MONDO term can be obsolete in MONDO but not in EFO? (excluding bugs of course). Aren't they supposed to mirror each other?

traitcuration/traits/datasources/dummy.py
joj0s: This is just for development purposes so that we can use the dummy admin credentials that are pinned in the Slack channel. 

I am manually adding admin@admin.com each deployment but I want to add a bash script that is executed upon deployment release and adds that user automatically.

I don't think the domain not existing is a problem, the sole purpose of those creds is to access the admin site, but if you think it would be better, I could change it to something like 'admin@example.com'

traitcuration/traits/datasources/dummy.py
joj0s: Just to clarify, by comment you mean a Python comment or a comment in the term label like *- INCORRECT* ?

traitcuration/traits/datasources/dummy.py
joj0s: About that, can a term be in EFO and the label change only in the parent ontology? And if so, how do I change the label?

traitcuration/traits/datasources/ols.py
joj0s: Yes, I agree, I will rethink that and merge those two function into a single calculation

traitcuration/traits/datasources/dummy.py
joj0s: Actually this is an interesting example I found. Here is term **Orphanet:98602** in EFO https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http://www.orpha.net/ORDO/Orphanet_98602 and in Orphanet https://www.ebi.ac.uk/ols/ontologies/ordo/terms?iri=http://www.orpha.net/ORDO/Orphanet_98602.

Which label would I keep in this situation?

traitcuration/traits/datasources/dummy.py
tskir: > I don't think the domain not existing is a problem

No, my concern is the opposite one, that the admin.com domain _does_ exist and hosts some real life webiste. IANA [recommends](https://www.iana.org/domains/reserved) that all such placeholder emails/addresses use the example.com domain, which is specifically reserved for this purpose. I agree this is not critical, just a good practice to follow

traitcuration/traits/datasources/dummy.py
tskir: Well... Originally I meant the Python comment, but now that you said it, adding “INCORRECT” to the label would also be useful :)

traitcuration/traits/datasources/dummy.py
tskir: Yes, that's curious. This is because, while the EFO maintains the original ontology CURIE (Orphanet:98602) on import, the term in EFO under that name now _belongs_ to EFO, so they can make arbitrary changes to it. Because our main ontology is EFO, in case of names conflict, EFO label should take precedence

traitcuration/traits/datasources/ols.py
tskir: Yeah, that's complicated. As you can see, in the example you reported above, the label is different in Orphanet and EFO for the same term, presumably because it was edited in EFO after import.

My best current understanding is that "importing" is more like "copying" the term from other ontology into EFO. The identifier, including the parent ontology's prefix, is maintained, but the contents can change. I think that even during regular import, they sometimes edit the title and description compared to how it was in the original ontology. And consequently, if something changes in the parent ontology (including the obsolete status), it will probably not be automatically reflected in EFO.

Or maybe they do try to keep them in sync, it is just not always possible due to bugs or complexity of it all. I don't know. I have to admit that even after working with ontologies for 1.5 years, I still do not now how _all_ of the import rules work. I guess at some point I will have to contact SPOT and to clarify this.

For now, let's adopt this approach (it's different from what I said in the starting comment of this thread):
* Treat EFO as the main deciding ontology for both the label and deciding the deleted/obsolete status
* Still always query the parent ontology and report (log) all cases where there is any discrepancy between EFO and the parent ontology

Does this sound all right to you?

joj0s: Sounds quite reasonable, yes

traitcuration/traits/datasources/dummy.py
tskir: OK, I'm going to resolve this conversation for now; but I've added a point to #58 to get rid of this in production

tskir (approved):

PR #82, opened 2020-08-18
Automatic mapping functionality and OxO queries

### LINKED ISSUE #44: Port the logic of the trait mapping pipeline
This is a follow-up ticket for #30. The "trait mapping pipeline" in the existing repository uses ZOOMA and OxO to attempt to _automatically_ map trait names to ontology terms, and outputs two files:
* Automatic mappings, which are considered of good quality and not manually reviewed (under current approach, ever)
* Mappings requiring manual curation, which are reviewed using spreadsheets.

The logic of that pipeline needs to be ported into the web app as well. One way this could work is that if a mapping is considered of good quality, it can be applied to the trait automatically, and its status can be moved directly into review (skipping the manual mapping part).

joj0s:
This PR adds mapping traits to term suggestions automatically, and creating new suggestions based on OxO queries. Pipeline logic:
1. If a term suggestion is of 'HIGH' confidence and is in EFO compatible ontologies, map it automatically to the trait.
2. If a term suggestion is an exact text match with the trait, map it automatically.
3.  If a term suggestion is of 'HIGH' confidence and is *8not** in EFO compatible ontologies, find its cross-refs using OxO, and create mapping suggestions of it, **but don't create a mapping**.

It can be tested by importing traits from ClinVar and then triggering a ZOOMA import.

Closes #44

### DIFF
--- a/traitcuration/traits/datasources/zooma.py
+++ b/traitcuration/traits/datasources/zooma.py
@@ -167,7 +167,7 @@ def find_automatic_mapping(trait, created_terms, high_confidence_term_iris):
         oxo_results = make_oxo_query([term_curie])
         for result in oxo_results['_embedded']['searchResults'][0]['mappingResponseList']:
             result_iri = make_ols_query(identifier_value=result['curie'],
-                                        ontology_id=ontology_id, identifier_type='curie')['iri']
+                                        ontology_id=ontology_id, identifier_type='obo_id')['iri']
             suggested_term = create_local_term(result_iri)
             created_terms.add(suggested_term)
             create_mapping_suggestion(trait, suggested_term)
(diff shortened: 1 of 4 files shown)


traitcuration/traits/datasources/zooma.py
```diff
@@ -29,26 +30,51 @@ def run_zooma_for_all_traits():
...
+
+    datasource_suggestion_list = get_suggestions_from_datasources(trait)
+    ols_suggestion_list = get_suggestions_from_ols(trait)
+
+    high_confidence_term_iris = list()  # A list of terms with 'HIGH' mapping confidence
+    final_suggestion_iri_set = set()  # A set of the term iris which are kept from the ZOOMA queries
+
+    for suggestion in datasource_suggestion_list + ols_suggestion_list:
+        if len(suggestion["semanticTags"]) > 1:
+            logger.warn("Suggestion with combined terms found! Skipping suggestion...")
```
tcezard, line 42: Can we log the actual suggestion that shows combined terms. it might be useful for later.

tcezard (approved):

PR #85, opened 2020-08-21
Maintainer feedback page and functionality

### LINKED ISSUE #73: Implement the page for submitting feedback to EFO
This should follow the UI designed in #72. On clicking the “Submit issue” button, the following sequence of events should take place:
1. Issue is created via GitHub API in the given repository
1. _Provided the previous action is successful,_ all terms which were part of the submission change their status from “needs (import/creation)” to “awaiting (import/creation)”

joj0s:
This PR adds a page to submit feedback to maintainers. The submission process includes creating a spreadsheet containing potential terms to create/import, and also creating a GitHub issue.

Some notes about this PR: 
- It is built on top of PR #77. To make the review process easier, I suggest using branch `ols_queries` as the base branch. PR #77 should be merged soon, and then I'll rebase this one.
- The Heroku deployment requires some configuration of GitHub and Google APIs to work. It should be up and running by the time this PR is reviewed, but if not, please report it in this PR's comments.

Closes #73 

This is deployed in http://trait-curation-2.duckdns.org/traits/


joj0s:
@tskir There is a problem with this PR that I found out while deploying it on Embassy. For some reason, when I deployed the app to Embassy, the OAuth window doesn't open in the user's browser, and there is only a prompt in the server with an OAuth link showing up. 

Even if I find a workaround, the problem I found out while researching this issue, is that as far as I understand, when a log in with the Sheets and Drive API is successful, the user's `client_id` and access token is saved permanently in the server, and the app uses those credentials to authenticate with Google Sheets pretty much permanently. I haven't been able to test it, but this might result in a user creating a spreadsheet with the creds of another user who authenticated before him.

The alternatives are two: I think the best approach is to create a service account for the app. A service account is a special type of Google account intended to represent a non-human user that needs to authenticate and be authorized to access data in Google APIs. I think it makes the most sense to create the sheets under this account and I wouldn't have to change the spreadsheet generation code at all, only the authentication section. However, I'm not sure if a service account can actually create sheets besides reading them. I will research this immediately.

The other option is to ditch the `gspread` library. The problem is that two other libraries that I found, seem to use the same authentication strategy, so the only alternative seems to be to use the Google API directly and not through an external library.

joj0s:
Update: Creation of a sheet with a service account is indeed successful. 
Here you can see it: https://docs.google.com/spreadsheets/d/1sgIYCPK6M4-7xMl-gDZiAqQLPjJerdO44xwI86EYXo8/

And the owner looks like this:
image

Should I go ahead and make the changes to the code (actually only a single line changes), and add instructions for obtaining service account credentials?

tskir:
Hrm. So it follows that the gspread library does not support multiple users or access keys? That's unfortunate. I was fine with just storing the credentials file on the server, but not with reusing it across different users.

Using a service account seems like a reasonable option; my only concern is who would then have admin control over that table? For example, if we wanted to delete it, how would we go about it?

traitcuration/static/js/github-issue.js
```diff
@@ -0,0 +1,95 @@
...
+  errorLabels = document.querySelectorAll(".field__error");
+  for (const label of errorLabels) label.innerHTML = "";
+
+  fields = document.querySelectorAll(".field");
+  for (field of fields) field.setCustomValidity("");
+
+  // Make sure the GitHub repo provided is valid and exists
+  try {
+    await axios
+    .get(`https://api.github.com/repos/${issue_repo.value}`, {
```
tskir, line 18: I feel a bit uneasy about making this GET request with an unsanitised user input (`${issue_repo.value}`). Is there a way we could do sanitisation in JS?

joj0s:
> Using a service account seems like a reasonable option; my only concern is who would then have admin control over that table? For example, if we wanted to delete it, how would we go about it?

You can specify that when creating the service account. For the one I just created, I gave everyone with an `ebi.ac.uk` domain access to admin operations with that account.

tskir (commented):
It looks good to me (see just one comment about input sanisiation). After we resolve this issue and also the one with gspread and service accounts, this PR can be merged

tskir:
> You can specify that when creating the service account. For the one I just created, I gave admin access to everyone with an ebi.ac.uk Google domain account.

I tried opening that spreadsheet from my EBI Google account and unfortunately I still have the "Move to bin" option greyed out.

I guess, if necessary, we could remove/administrate the spreadsheets using the service account credentials directly. In this case, we need to make sure that for any given installation the service account credentials are persistent. Looking at [gspread documentation](https://gspread.readthedocs.io/en/latest/oauth2.html#service-account), this should be possible like this:

```
gc = gspread.service_account(filename='path/to/the/downloaded/file.json')
```

joj0s:
> I tried opening that spreadsheet from my EBI Google account and unfortunately I still have the "Move to bin" option greyed out.

I gave myself access and tried it too with the same result, that is weird, I will look into it.

> `gc = gspread.service_account(filename='path/to/the/downloaded/file.json')`

I am fairly certain that this option just changes the default path that gspread looks for to find the `service_account.json` file. If we omit that, the service acount creds will still persist in the default folder, which is `~/.config/gspread/service_account.json`.

joj0s:
A good alternative would be to give 'owner' permission to specific people for a spreadsheet using gspread, when I create the spreadsheet. We could follow a similar approach where I give 'owner' permission to everyone with a `ebi.ac.uk` domain account, or to the app's admins.

Reference: https://gspread.readthedocs.io/en/latest/api.html#gspread.Client.insert_permission

tskir:
> I am fairly certain that this option just changes the default path that gspread looks for to find the service_account.json file. If we omit that, the service acount creds will still persist in the default folder, which is ~/.config/gspread/service_account.json.

Ah, I see. The location doesn't matter that much, I just hope that it will indeed _persist,_ i.e., that it will not unnecessarily download a new auth JSON every time, including on app restarts.

> A good alternative would be to give 'owner' permission to specific people for a spreadsheet using gspread, when I create the spreadsheet. We could follow a similar approach where I give 'owner' permission to everyone with a ebi.ac.uk domain account, or just to the app's admins.

If we could grant ownership access to all app admins, that would be great. Of course, provided that this actually works, since it currently appears not to, at least regarding sheet deletion :-)

joj0s:
> If we could grant ownership access to all app admins, that would be great. Of course, provided that this actually works, since it currently appears not to, at least regarding sheet deletion :-)

So, I am adding 'edit' access to anyone and owner access to EBI acccounts, does that sound good?

tskir:
> So, I am adding 'edit' access to anyone and owner access to EBI acccounts, does that sound good?

Edit access to anyone is good; and owner access should be restricted to app admins, not all EBI accounts on the whole (especially considering that in the future this might have use cases outside of EBI)

> @tskir did you get an email notifying that you have ownership of the latest spreadsheet?

No, I'm afraid I didn't get any emails regarding this

joj0s:
> No, I'm afraid I didn't get any emails regarding this

That's really curious, using this code, I got ownership for the latest spreadsheet and I got an email notifying me. Can you confirm I entered your email correctly?

```python
gc.insert_permission(sheet.id, value = 's3apostol@gmail.com', perm_type = 'user', role = 'owner')
gc.insert_permission(sheet.id, value = 'ktsukanov@ebi.ac.uk', perm_type = 'user', role = 'owner')
```

joj0s:
I updated the sheet creation process. I used batch update to insert all values at once per worksheet. So now I am making four requests in total: 
- One to add different formatting to the 'needs_import" worksheet headers
- One to add all the mapping information to the 'needs_import' worksheet
- One to add different formatting to the 'needs_creation worksheet headers
- One to add all the mapping information to the 'needs_creation worksheet

I also updated the deployment of the 'feedback' branch in http://trait-curation-2.duckdns.org/

traitcuration/static/js/github-issue.js
joj0s: So regarding this one, if you are referring to converting to HTML encoding to escape non standard symbols I could write something like this:
``` js
function sanitize(string) {
  const map = {
      '&': '&amp;',
      '<': '&lt;',
      '>': '&gt;',
      '"': '&quot;',
      "'": '&#x27;',
      "/": '&#x2F;',
  };
  const reg = /[&<>"'/]/ig;
  return string.replace(reg, (match)=>(map[match]));
}```

However, since all I am making is a GET request to an external API, I'm not sure if it is unsafe. Should I go ahead and use the above `sanitize` function?

joj0s:
@tskir I added transferring of ownership upon successful spreadsheet creation. When you can, please go to http://trait-curation-2.duckdns.org/traits/ and try to create an issue and confirm whether you have been made owner of the generated spreadsheet.

traitcuration/static/js/github-issue.js
tskir: Since this is a request to an external API made by the user, I agree this is not unsafe in terms of XSS or similar security risk; still, I'd rather avoid accidental nonsensical API requests in case a user badly mistypes something in a field.

To apply this `sanitize` function would a good idea (in the future, it can also be reused throughout the project for similar purposes). A possible alternative is to simply validate GitHub repository name using a regular expression. I would be fine with either of those solutions.

tskir (commented):
Regarding the batch operations: that's great! I imagine the performance should be significantly higher now that we only make a handful requests.

Regarding the spreadhseet ownership approach: unfortunately I can't test it, because for some reason there's no feedback form on the Feedback page. Do you know why this might be? 
image

joj0s:
> Regarding the spreadsheet ownership approach: unfortunately I can't test it, because for some reason there's no feedback form on the Feedback page. Do you know why this might be?

Yes, this is because I am hiding the form when there are no terms to import. I went ahead and reviewed one from the dummy data so that you can test this

tskir:
> Yes, this is because I am hiding the form when there are no terms to import.

This doesn't seem right: the form should be shown when there is at least one term to _either_ import or create. While rare for our use case, I can definitely imagine a situation where someone would want to request creation of 10 terms and import of none.

If this can be fixed quickly, then we can do this as part of this PR; otherwise we can create a follow-up issue

joj0s:
> If this can be fixed quickly, then we can do this as part of this PR; otherwise we can create a follow-up issue

This should now be fixed.

tskir:
@joj0s The submission form looks great and the GitHub login works. Unfortunately, when I try to actually do the submission, I get redirected to: `http://localhost:8000/traits/github/callback/?code=[some token]`, which of course doesn't work because I don't have anything deployed on my local machine. It's probably a deployment configuration issue, could you take a look into that when you have time?

In the meanwhile, I'm going to merge this PR because the code changes themselves all look good.

Issue #34, opened 2020-06-06
Set up automated periodic updates of external datasources


tskir:
Related to #32. In addition to being able to trigger imports manually, there should also be automated periodic updates.

Care needs to be taken (in both cases) to avoid any race conditions when a user tries to change something in the same moment when an update operation is being performed.

joj0s:
@tcezard  I can confirm task automation is successful. What tasks should I schedule at this point and when? Available background tasks right now are ZOOMA suggestion imports and ClinVar imports, plus OLS imports when PR #77 is merged.

joj0s:
Also, there exists a very useful Django package that lets the user schedule tasks through the Django Admin with a very friendly interface.
![image](https://user-images.githubusercontent.com/36301625/90748395-d0ad7d00-e2da-11ea-9d6d-c06e5593a7bf.png)
So maybe it would be preferable that I don't hardcode any scheduled task, but I let the admins do that?

tcezard:
It would make a lot more sense not to have to reinvent the wheel and use this module. Can we predefine the tasks ? Does it run python function or command line executable ?

tskir:
> By using the mentioned django_celery_beat module, we can configure the hardcoded task schedule via the Django Admin, which will save us the time of creating our own configuration page for that. However if the server restarts, the hardcoded schedule will again be used.

I'd say this solution is not very good, as it lacks settings persistence.

> The alternative, while still using the module, is to not hardcode anything, but instead have the admins define the scheduled tasks via the Django Admin. Again, the tasks will be already defined, and all the admins will have to do, is to select the task to run from a dropdown menu, and when to run them via a form.

This sounds somewhat better, but would it also just reset after every server restart?

Overall, I'd say that while it would be nice to have some sort of a UI to configure periodic import schedule, at this point this is not completely necessary. This is a setting which will be changed very very rarely and only by the app's admins. So even hardcoding that in settings.py sounds acceptable to me, if there are no other good (persistent) solutions

tskir:
Also I wonder if we could use the admin page in issue #50 for that. There, we already have some configurable parameters, which will have to be persistent. For example, we have a configurable ZOOMA base URL. Could we just add a conifgurable ZOOMA query schedule field next to that? Or does it _have_ to be hardcoded in settings.py?

joj0s:
> This sounds somewhat better, but would it also just reset after every server restart?

[...]

PR #87, opened 2020-08-26
Datasources page

### LINKED ISSUE #83: Implement the 'data sources' page
Implementation of the page designed in issue #32.

joj0s:
This PR adds a page to display external data sources, along with their latest import progress and status. Closes #83 

Compared to the UI design, I removed the 'Next scheduled import' section due to the open discussion in issue #34.

As usual, since this PR depends on changed introduced in PR #85, it is better to switch to `feedback` as the base branch for the review process and when that PR is merged I will rebase.


traitcuration/static/js/datasources.js
```diff
@@ -9,3 +63,34 @@ function makeDatasourceRequest(selectedDatasource) {
...
+  progressBarElement.style.backgroundColor = '#68a9ef';
+  progressBarElement.style.width = progress.percent + "%";
+  progressBarMessageElement.innerHTML = progress.current + ' traits of ' + progress.total + ' processed...';
+}
+
+function traitImportProgress (progressBarElement, progressBarMessageElement, progress){
+  progressBarElement.style.backgroundColor = '#68a9ef';
+  progressBarElement.style.width = progress.percent + "%";
+  progressBarMessageElement.innerHTML = progress.current + ' MB of ' + progress.total + ' MB downloaded...';
+}
```
tskir, line 91: The blocks for OLS, ZOOMA and ClinVar look very similar (here and also above in the same file). It looks fine for now, but when in the future we add more trait/data sources, we might have to think about how to condense them into a single configurable method.

traitcuration/traits/utils.py
```diff
@@ -38,6 +39,55 @@ def parse_request_body(request):
...
+        clinvar_task_id = latest_clinvar_import.task_id
+    else:
+        latest_import_date = None
+        clinvar_task_id = "None"
+    context['clinvar_task_id'] = clinvar_task_id
+    clinvar_dict = {'id': 'clinvar', 'title': 'ClinVar',
+                    'latest_import_date': latest_import_date}
+    trait_sources.append(clinvar_dict)
+
+    context['trait_sources'] = trait_sources
```
tskir, line 88: Same comment about (mild) code block duplication applies here. Right now it's probably not worth the effort to refactor that, but once we have more sources, it will become necessary

tskir (commented):
Looks good to me! I left only two minor comments about some mild code duplication for OLS/ZOOMA/ClinVar cases. If you have an idea how to resolve that reasonably easily, then do go ahead; otherwise, we can just leave that as is for now and only worry about that once we have a significantly larger number of data sources

tskir:
(I've also changed PR base to `feedback` for easier discussions, this should be set to `master` before merging)

traitcuration/traits/utils.py
joj0s: As with the above, it is kinda tricky to condense that, I will rethink that, but I think we should merge this PR and add a follow up issue about that.

tskir: Agreed

tskir (approved):
In terms of code changes, everything's good here. I would also like to look at the deployment once we resolve the issue with feedback form missing in #85

tskir:
(Also note that this branch has now accumulated some merge conflicts)

tskir:
@joj0s Everything looks good here, I'm going to go ahead and merge this one as well. I'll test this functionality at the same time as functionality from other pending PRs

PR #94, opened 2020-09-04
Update ClinVar parsing script to meet new specifications

### LINKED ISSUE #84: Multiple trait names on single 'PhenotypeList" columns during latest ClinVar import
During the latest ClinVar import I encountered some weird behaviour. There were multiple trait names for single records, seperated by '|', looking like this: 
image

In all past releases, they were separated with semicolons.

I imagined it was a bug with the parsing script, however when I downloaded the latest ClinVar relese manually, it turns out the PhenotypeList column contained a lot of those records:
image

Is this behaviour to be expected?

joj0s:
This PR includes three small changes in the ClinVar parsing script based on the updates mentioned in #84 
- Replaced semicolon(';') with pipe('|') as the seperating character for phenotypes and RCV
- Added a check to remove '-' entries in the PhenotypeList column
- Added a check to skip rows that contain more than 5 phenotypes

Closes #84 

A working deployment of this PR is available here: https://trait-curation-pr-94.herokuapp.com/traits/

### DIFF
--- a/traitcuration/traits/datasources/clinvar.py
+++ b/traitcuration/traits/datasources/clinvar.py
@@ -61,8 +61,12 @@ def parse_trait_names_and_source_records():
         for row in itertools.islice(reader, NUMBER_OF_RECORDS):
             # For every row, get its allele_id and all its rcv_accessions and phenotypes
             row_alleleid = (row['#AlleleID'])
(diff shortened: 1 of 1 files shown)


tskir:
Looks good to me, thanks!

PR #95, opened 2020-09-04
Add external config file and conditional project settings

### LINKED ISSUE #86: Remove the autogenerated secret key from settings.py
The `traitcuration/settings.py` file [contains](https://github.com/EBIvariation/trait-curation/blob/c1d467472442758d28aea5983c55c118a1b02e56/traitcuration/settings.py#L24) an automatically generated secret key. While we don't use it anywhere, it should still be removed to avoid confusion.

joj0s:
This PR adds the ability to read settings and credentials from an external `config.yaml` file, and to adjust the project settings based on a `DJANGO_ENV` environment variable. 

---
Added by @tskir: Closes #86.


tskir:
Everything looks good, I don't have any comments. I will test this functionality, along with the deployment instructions, in #96.

